PR #4, opened 2019-10-07
Kubernetes example and redux

dvaldivia:
scripts to provision a kubernetes cluster locally using `kind` which runs the nodes inside containers.

build the modified

with the script `create-kind.sh`  it will
* create the kind cluster
* configure dashboard
* create folder for the simulated tenant
* install the tenant


k8s/add-volume.sh
```diff
@@ -0,0 +1,21 @@
+TENANT_NAME="${1:-tenant-1}"
```
Alevsk, line 1: can you ddd `#!/bin/bash` at the beginning of the file just for convention.

k8s/create-kind.sh
```diff
@@ -0,0 +1,16 @@
+echo "Provisioning Kind"
```
Alevsk, line 1: can you ddd #!/bin/bash at the beginning of the file just for convention.

k8s/deployments/tenant1-env.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tenant-1-env
+  labels:
+    app: tenant-1
+data:
+  MINIO_ACCESS_KEY: "minio"
```
Alevsk, line 8: For secrets in the future we can use a solution like `envsubst` 
 https://stackoverflow.com/questions/56003777/how-to-pass-environment-variable-in-kubectl-deployment#answer-56009991

Alevsk (approved):

cesnietor:
are **k8s/deployments/tenant-1.yaml**  and **k8s/deployments/tenant1-env.yaml** planned to be used?

dvaldivia:
@cesnietor yes! there's a dependency between the two, we need the secret first and then the deployment itself

PR #5, opened 2019-10-08
V3 deployment strategy

dvaldivia:
This includes a new deployment strategy that uses shared `storage-cluster` services to identify  the nodes

the `create-kind.sh` script now prepares the nodes for `2 tenants`

the deployment strategy is basically in order, so in the `deployment/v3/` you'll find a prefix that ensures `kubectl apply -f` applies everything int he right order


dvaldivia:
here's a diagram of how it looks
Screen Shot 2019-10-08 at 1 36 45 PM

dvaldivia:
Diagram of how this deployment strategy looks

Screen Shot 2019-10-08 at 1 45 59 PM

cesnietor (approved):
Approved after live review

PR #6, opened 2019-10-09
Provisiong storage cluster with go code

dvaldivia:
Provisioning the `v3` strategy using the kubernetes API.

open `cluster/cluster.go` and put your local k8s serviceaccount token on line 47 where it says `<YOUR_TOKEN>` 
build the project then run `./m3 cluster sc add`

This will create the storage cluster and two tenants as before, we can use this as a basis.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ WORKDIR /
 RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-s" -a -installsuffix cgo -o main ./go/src/github.com/minio/m3/
 
 FROM scratch
-MAINTAINER Daniel Valdivia "daniel@min.io"
+MAINTAINER MinIO Development "dev@min.io"
 EXPOSE 10105
 
 COPY --from=0 /main    .
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ADD . /go/src/github.com/minio/m3/
 WORKDIR /go/src/github.com/minio/m3/
 WORKDIR /
 
-RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-s" -a -installsuffix cgo -o m3 ./go/src/github.com/minio/m3/
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-w -s" -a -installsuffix cgo -o m3 ./go/src/github.com/minio/m3/
 
 FROM scratch
 MAINTAINER MinIO Development "dev@min.io"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM golang:1.13.1
+
+#ENV GO111MODULE=on
+#ENV GOPROXY=https://proxy.golang.org
+#ENV GOPROXY=direct
+
+ADD go.mod /go/src/github.com/minio/m3/go.mod
+ADD go.sum /go/src/github.com/minio/m3/go.sum
+WORKDIR /go/src/github.com/minio/m3/
+# Get dependancies - will also be cached if we won't change mod/sum
+RUN go mod download
+
+ADD . /go/src/github.com/minio/m3/
+WORKDIR /go/src/github.com/minio/m3/
+WORKDIR /
+
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-s" -a -installsuffix cgo -o main ./go/src/github.com/minio/m3/
+
+FROM scratch
+MAINTAINER Daniel Valdivia "daniel@min.io"
+EXPOSE 10105
+
+COPY --from=0 /main    .
+
+CMD ["/main"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,5 @@
 FROM golang:1.13.1
 
-#ENV GO111MODULE=on
-#ENV GOPROXY=https://proxy.golang.org
-#ENV GOPROXY=direct
-
 ADD go.mod /go/src/github.com/minio/m3/go.mod
 ADD go.sum /go/src/github.com/minio/m3/go.sum
 WORKDIR /go/src/github.com/minio/m3/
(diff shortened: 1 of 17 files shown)


Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ADD go.sum /go/src/github.com/minio/m3/go.sum
+WORKDIR /go/src/github.com/minio/m3/
+# Get dependancies - will also be cached if we won't change mod/sum
+RUN go mod download
+
+ADD . /go/src/github.com/minio/m3/
+WORKDIR /go/src/github.com/minio/m3/
+WORKDIR /
+
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-s" -a -installsuffix cgo -o main ./go/src/github.com/minio/m3/
```
harshavardhana:
```suggestion
RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-w -s" -a -installsuffix cgo -o main ./go/src/github.com/minio/m3/
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM scratch
+MAINTAINER Daniel Valdivia "daniel@min.io"
```
harshavardhana:
```suggestion
MAINTAINER MinIO Development "dev@min.io"
```

harshavardhana: Avoid personal names wherever we can

cluster/cluster.go
```diff
@@ -0,0 +1,373 @@
...
+				Spec: mainPodSpec,
+			},
+		},
+	}
+
+	res, err := clientset.ExtensionsV1beta1().Deployments("default").Create(&deployment);
+	if err != nil {
+		panic(err.Error())
+	}
+	fmt.Println("done creating storage cluster deploymenty ")
```
harshavardhana, line 372: Is there a kubernetes specific logging helper we can use ?

cluster/cluster.go
```diff
@@ -0,0 +1,373 @@
...
+		return "m3cluster-worker4"
+	default:
+		return "m3cluster-worker"
+	}
+}
+
+func getDisks(storageClusterNum string, hostNum string) []string {
+	switch storageClusterNum {
+	case "2":
+		return []string{"/mnt/disk5", "/mnt/disk6", "/mnt/disk7", "/mnt/disk8"}
```
harshavardhana, line 270: Is the prefix here /mnt/disk going to be constant ?

cluster/cluster.go
```diff
@@ -0,0 +1,373 @@
...
+	default:
+		return []string{"/mnt/disk1", "/mnt/disk2", "/mnt/disk3", "/mnt/disk4"}
+	}
+
+}
+
+const MaxNumberDiskPerNode = 4
```
harshavardhana:
```
const (
)
``` 

Group then together..

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+#ENV GO111MODULE=on
+#ENV GOPROXY=https://proxy.golang.org
+#ENV GOPROXY=direct
+
+ADD go.mod /go/src/github.com/minio/m3/go.mod
+ADD go.sum /go/src/github.com/minio/m3/go.sum
+WORKDIR /go/src/github.com/minio/m3/
+# Get dependancies - will also be cached if we won't change mod/sum
+RUN go mod download
```
harshavardhana, line 7: We don't need to do this. This is automatic

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.13.1
+
+#ENV GO111MODULE=on
+#ENV GOPROXY=https://proxy.golang.org
+#ENV GOPROXY=direct
+
+ADD go.mod /go/src/github.com/minio/m3/go.mod
+ADD go.sum /go/src/github.com/minio/m3/go.sum
```
harshavardhana, line 4: You can add .dockerignore such that to avoid `.git` etc.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.13.1
+
+#ENV GO111MODULE=on
+#ENV GOPROXY=https://proxy.golang.org
```
harshavardhana: This is default with go1.13

Dockerfile
harshavardhana: Why not add -o m3 and then use

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-s" -a -installsuffix cgo -o main ./go/src/github.com/minio/m3/
+
+FROM scratch
+MAINTAINER Daniel Valdivia "daniel@min.io"
+EXPOSE 10105
+
+COPY --from=0 /main    .
+
+CMD ["/main"]
```
harshavardhana:
```suggestion
CMD ["/m3"]
```

harshavardhana (changes requested):

cluster/cluster.go
dvaldivia: anything that goes to std/out becomes a log which can be collected by k8s, these log lines are temporary

Dockerfile
dvaldivia: what does `-w` do?

cluster/cluster.go
dvaldivia: not really, this should be a smarter function that knows how many disks are available and return the proper array, I guess the prefix is up for grabs, what do you think is the best mount point for the disks?

Dockerfile
dvaldivia: I know, but I wanted cached as a layer since during development we most likely will be building the docker and pushing it to k8s to test

Dockerfile
dvaldivia: found out https://stackoverflow.com/questions/22267189/what-does-the-w-flag-mean-when-passed-in-via-the-ldflags-option-to-the-go-comman

Alevsk:
I was able to run this code locally and the tenant provisioning was working fine 👍

Alevsk (approved):
👍

PR #23, opened 2019-10-11
remove some unnecessary code copied from mc

aead:
This commit removes some unnecessary code and
tries to clean some parts up. However, there are
still more things to remove (e.g. `S3 client
factory - that's not Go at all.)


dvaldivia:
yes, initially we pulled code from `mc` to be able to use `madmin`, I'm wondering if we should instead add the support for madmin on `minio-go` and use that instead?

harshavardhana:
> yes, initially we pulled code from `mc` to be able to use `madmin`, I'm wondering if we should instead add the support for madmin on `minio-go` and use that instead?

madmin requires server side deps that's why its kept inside minio server @dvaldivia - once we stabilize the madmin we can move to minio-go or a separate go package with semantic versioning. Unfortunately, right now too many things are changing.

aead:
@dvaldivia At the current state we don't need behave the same way as the minio server. As long as the general m3 sdk API's are well-defined we can change e.g. logging later on easily.

dvaldivia:
@harshavardhana yeah, let's use madmin in the mean while
@aead sent you a PR with some changes to address the build error

dvaldivia (approved):

PR #26, opened 2019-10-14
Track tenants on db

dvaldivia:
This adds support for db using postgres as the backend.

To setup a brand new m3 on a kubernetes cluster, the operator would need to do:

`m3 setup` - Setups m3 on it's own k8s namespace and installs postgres there
`m3 setup db` - Optional command that runs the migrations
`m3 cluster sc add --name OPTIONALNAME` - Adds a new Storage Cluster
`m3 tenant add --name TENANT --short_name SHORTNAME` - Adds a new tenant to the tenants table, and provisions it on a randomly selected storage cluster

After each `m3 tenant add` give some time for the deployments to settle (about 40 seconds) before the tenant becomes avaiable.

It's important `m3 cluster sc add` is run at least once, ideally on setup we would allocate a storage cluster only for private needs (like access logs and other data) but for now, let's do it manually.


cluster/cluster.go
```diff
@@ -175,30 +174,29 @@ func CreateSCHostService(storageClusterNum string, hostNum string, prefix *strin
...
 	if err != nil {
-		panic(err.Error())
+		return err
 	}
-	fmt.Println(res.String())
-
+	return nil
 }
 
 //Creates a the "secrets" of a tenant, for now it's just a plain configMap, but it should be upgraded to secret
```
aead:
```suggestion
// Creates a the "secrets" of a tenant, for now it's just a plain configMap, but it should be upgraded to secret
```

cluster/cluster.go
```diff
@@ -175,30 +174,29 @@ func CreateSCHostService(storageClusterNum string, hostNum string, prefix *strin
...
-func CreateTenantConfigMap(tenantShortName string) {
+func CreateTenantConfigMap(tenant *Tenant) {
```
aead: `CreateTenantConfigMap` should return an error instead of panic'ing / printing errors.

cluster/cluster.go
```diff
@@ -207,11 +205,13 @@ func CreateTenantConfigMap(tenantShortName string) {
...
 	}
 
-	res, err := clientset.CoreV1().ConfigMaps("default").Create(&configMap);
+	res, err := clientset.CoreV1().ConfigMaps("default").Create(&configMap)
 	if err != nil {
 		panic(err.Error())
 	}
-	fmt.Println(res.String())
+	if res.Name == "" {
+		fmt.Println("Error adding config map")
```
aead: Return an error

cluster/cluster.go
```diff
@@ -244,19 +244,17 @@ func CreateTenantService(tenantName string, tenantPort int32, storageClusterNum
...
-	fmt.Println(res.String())
+	fmt.Printf("done creating tenant service for tenant %s \n", res.Name)
 
 }
 
-type Tenant struct {
-	Name              string
-	Port              int32
-	StorageClusterNum string
+type SCTenant struct {
```
aead: What is an `SCTenant` (go doc :wink: )

cluster/cluster.go
```diff
@@ -244,19 +244,17 @@ func CreateTenantService(tenantName string, tenantPort int32, storageClusterNum
...
+	Name string
+	Port int32
```
aead: What is a `Port`?

cluster/cluster.go
```diff
@@ -290,7 +288,7 @@ const (
 )
 
 //Creates a service that will resolve to any of the hosts within the storage cluster this tenant lives in
-func CreateDeploymentWithTenants(tenants []Tenant, storageClusterNum string, hostNum string, prefix *string) {
+func CreateDeploymentWithTenants(tenants []*StorageClusterTenant, storageClusterNum string, hostNum string, prefix *string) {
```
aead: This function should return an error instead of panic'ing

aead:
```suggestion
func CreateDeploymentWithTenants(tenants []*StorageClusterTenant, storageClusterNum, hostNum, prefix *string) {
```

cluster/cluster.go
```diff
@@ -290,7 +288,7 @@ const (
 )
 
 //Creates a service that will resolve to any of the hosts within the storage cluster this tenant lives in
```
aead, line 291: What are the parameters for?

cluster/cluster.go
```diff
@@ -394,66 +392,47 @@ func CreateDeploymentWithTenants(tenants []Tenant, storageClusterNum string, hos
...
+	ch := make(chan error)
 	go func() {
 		defer close(ch)
-		// create the tenant folder on each node via job
-		config := getConfig()
-		// creates the clientset
-		clientset, err := kubernetes.NewForConfig(config)
-		if err != nil {
-			panic(err.Error())
+		if tenant == nil || sc == nil {
```
aead, line 406: Is a `nil` tenant / sc expected? Shouldn't that be treated as error?

cluster/context.go
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
+
+import (
+	"context"
+	"database/sql"
+)
+
+type Context struct {
```
aead, line 27: What is a `Context`?

cluster/context.go
```diff
@@ -0,0 +1,32 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
```
aead, line 17: Insert new line to separate license header from go doc:
 ```
// You should have received a copy of the GNU Affero General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

package cluster
```

File: cluster/context.go [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"database/sql"
+)
+
+type Context struct {
+	*sql.Tx
+	Main *context.Context
+}
+
+// NewContext creates a new context.
```
aead: That comment is somewhat redundant - when/why should we use a context?

cluster/sessions.go
```diff
@@ -0,0 +1,101 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
aead, line 15: Insert new line to separate license header from go doc:
 ```
// You should have received a copy of the GNU Affero General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

package cluster
```

File: cluster/sessions.go [hunk 13]
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"database/sql"
+	"fmt"
+	_ "github.com/lib/pq"
+	"log"
+	"os"
+	"sync"
+)
+
+type singleton struct {
```
aead: This struct seems to be not necessary. Maybe that can be just a `var dbConn *sql.DB` ?

cluster/sessions.go
```diff
@@ -0,0 +1,101 @@
...
+		db := <-ConnectToDb()
+
+		instance = &singleton{
+			Db: db,
+		}
+	})
+	return instance
+}
+
+func ConnectToDb() chan *sql.DB {
```
aead: This function should take a context (for cancellation) and return a `*sql.DB`. For example the current code would block forever in case of an error.
Maybe we can use something like:
```
ConnectToDb(ctx context.Context) (*sql.DB, error) {
      select {
            case // try to connect to DB
            case // abort when context is canceled 
      }
}
```

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
```
aead, line 17: Insert new line to separate license header from go doc:
 ```
// You should have received a copy of the GNU Affero General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

package cluster
```

File: cluster/storage-cluster.go [hunk 16]
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+type StorageCluster struct {
```
aead, line 26: What is a storage cluster?

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	Id   int32
```
aead: Are we "limited" to 4 byte IDs? - Shouldn't that be a string?

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	Name *string
```
aead, line 28:
```suggestion
	Name string
```

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+type StorageClusterResult struct {
```
aead, line 32: What is a `StorageClusterResult`?

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	*StorageCluster
+	Error error
+}
+
+type AddStorageClusterResult struct {
```
aead: What is a `AddStorageClusterResult`? - Also why do we need a different type compared to `StorageClusterResult` that has no other fields. If really necessary maybe `type AddStorageClusterResult StorageClusterResult`

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	*StorageCluster
+	Error error
+}
+
+// Creates a storage cluster in the DB
+func AddStorageCluster(scName *string) chan AddStorageClusterResult {
```
aead:
```suggestion
func AddStorageCluster(scName string) chan AddStorageClusterResult {
```

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+				Id:   id,
+				Name: name,
+			},
+		}
+
+	}()
+	return ch
+}
+
+func GetListOfTenantsForSCluster(ctx *Context, sc *StorageCluster) chan []*StorageClusterTenant {
```
aead, line 134: What is the purpose of `GetListOfTenantsForSCluster`

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+				ServiceName:      serviceName,
+				StorageClusterId: sc.Id})
+
+		}
+		ch <- tenants
+	}()
+	return ch
+}
+
+type StorageClusterTenant struct {
```
aead, line 183: What is a `StorageClusterTenant`?

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	*Tenant
+	StorageClusterId int32
```
aead: Are we limited to 4 bytes IDs? - Shouldn't that be a string?

cluster/storage-cluster.go
```diff
@@ -0,0 +1,250 @@
...
+	Port             int32
+	ServiceName      string
+}
+
+type StorageClusterTenantResult struct {
```
aead, line 191: What is a `StorageClusterTenantResult`?

cluster/tenants.go
```diff
@@ -0,0 +1,133 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
```
aead, line 17: Insert new line to separate license header from go doc:
 ```
// You should have received a copy of the GNU Affero General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

package cluster
```

**Review by aead — CHANGES_REQUESTED** [review]

File: cluster/cluster.go (hunk 5 above)
Comment (dvaldivia) [reply cluster/cluster.go]: This was a unused struct so I've removed it :P

File: cluster/storage-cluster.go (hunk 24 above)
Comment (dvaldivia) [reply cluster/storage-cluster.go]: ideally the storage cluster id is a sequential number (ie sc-1, sc-2, sc-3) because it's convenient to use for auto-naming the storage clusters, so I hope it's enough.

File: cluster/context.go (hunk 9 above)
Comment (dvaldivia) [reply cluster/context.go:27]: An application wide context that holds the a transaction, in case anything goes wrong during the business logic execution, database objects can be rolled back.

File: cluster/sessions.go (hunk 14 above)
Comment (dvaldivia) [reply cluster/sessions.go]: This code has a fatal clause, since if there's not DB we wouldn't be able to do much, but yeah, i want to start using the context around to control cancellations, node that I `defer close(ch)` at the beginning of the function to avoid the block. I added the select clause.

File: cluster/sessions.go (hunk 13 above)
Comment (dvaldivia) [reply cluster/sessions.go]: This way we can hold multiple stateful connections.

File: cluster/storage-cluster.go (hunk 18 above)
Comment (dvaldivia) [reply cluster/storage-cluster.go:28]: Storage cluster name is optional, so the field may be `NULL` on the database.

File: cluster/storage-cluster.go (hunk 20 above)
Comment (dvaldivia) [reply cluster/storage-cluster.go]: my mistake xD didn't notice I had the same struct duplicate hahaha I'm glad we have code reviews.

File: cluster/setup.go [hunk 27]
```diff
@@ -0,0 +1,185 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
```
cesnietor, line 17: Insert new line to separate license header from go doc:

cluster/cluster.go
```diff
@@ -79,12 +80,12 @@ func ListPods() {
 }
 
 //Creates a headless service that will point to a specific node inside a storage cluster
-func CreateSCHostService(storageClusterNum string, hostNum string, prefix *string) {
+func CreateSCHostService(storageClusterNum string, hostNum string, prefix *string) error {
```
aead:
```suggestion
func CreateSCHostService(storageClusterNum string, hostNum string, prefix string) error {
```

cluster/cluster.go
```diff
@@ -397,71 +387,54 @@ func CreateDeploymentWithTenants(tenants []Tenant, storageClusterNum string, hos
...
-			CreateDeploymentWithTenants(tenants, "1", fmt.Sprintf("%d", i), nil)
-			// wait for the deployment to come online
+
+		CreateTenantService(scTenantResult.StorageClusterTenant)
+
+		// call for the storage cluster to refresh
+		err := <-ReDeployStorageCluster(ctx, sc)
+		if err != nil {
+			ch <- err
+			return
```
aead:
```suggestion
```

cluster/context.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"database/sql"
+)
+
+// An application wide context that holds the a transaction, in case anything
+// goes wrong during the business logic execution, database objects can be
+// rolled back.
+type Context struct {
+	*sql.Tx
+	Main *context.Context
```
aead: You are holding a pointer to an interface here. Is that on purpose?! If so, what is the reasoning. In general *do not* use interface indirection.

aead:
```suggestion
	Main context.Context
```

cluster/context.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Creates a new `Context` given an initial transaction and `context.Context`
+// to control timeouts and cancellations.
+func NewContext(tx *sql.Tx, ctx *context.Context) *Context {
```
aead:
```suggestion
func NewContext(tx *sql.Tx, ctx context.Context) *Context {
```

cluster/sessions.go
aead: Then I suggest `dbConnPool` (or similar) instead of `singleton`. Names should indicate the purpose of a thing.

cluster/setup.go
```diff
@@ -0,0 +1,186 @@
...
+	resDeployment, err := clientset.ExtensionsV1beta1().Deployments(m3SystemNamespace).Create(&deployment)
+	if err != nil {
+		panic(err.Error())
+	}
+	fmt.Println("done creating postgres deployment ")
+	fmt.Println(resDeployment.String())
+
+}
+
+func RunMigrations() {
```
aead, line 179: What migrations will be run by this function?

cluster/storage-cluster.go
aead: So, your saying that this struct may get marshaled into a SQL table entry? And this marshaling treats a nil pointer as `NULL` but `""`(empty string) not? If so, why don't we use the null-able types in the `sql` package? E.g. https://golang.org/pkg/database/sql/#NullString.

Also, if this (using `*string`) is necessary than we need comments explaining this since that is not obvious AFAICS.
E.g. `Name *string // Optional cluster name - nil pointer maps to NULL on DB`

aead (changes requested):

cluster/cluster.go
dvaldivia: I had forgotten to push some updates, I removed prefix all together

cluster/context.go
dvaldivia: I kind of want to have a traditional context attached to our application specific context so we can still handle cancellations and timeouts, this was the only way I could think of it, would you recommend I rather hold the instance instead of a pointer?

cluster/setup.go
dvaldivia: Added some documentation, it will run all the migrations, `go-migrate` keeps track off which migrations have been already applied to the database

aead:
@dvaldivia Is it okay to rebase this PR on master after #29 is merged to run it through the static lintings checks?

dvaldivia:
@aead sure

cesnietor:
Can you add the `$ m3 setup ...` commands to the readme?

cluster/setup.go
```diff
@@ -0,0 +1,191 @@
...
+	fmt.Println(res.String())
+
+	configMap := v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "postgres-env",
+			Labels: map[string]string{
+				"app": "postgres",
+			},
+		},
+		Data: map[string]string{
```
Alevsk, line 113: I think for postgres credentials using a kubernetes secret instead of kubernetes configMap is better but its good for now, we can change that later.

Alevsk (approved):
👍

cluster/setup.go
dvaldivia: Yes, we have to move to secrets eventually, we can do that on the next PR

aead (approved):

PR #43, opened 2019-10-18
Load balancer

Alevsk:


k8s/load-balancer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM nginx:latest
```
dvaldivia: I think we can use pure nginx without the need of our own image

Alevsk:
@aead Hi Andreas, I'll like your opinion and ideas on this :), this PR build the nginx.config file being used by the nginx to resolve the tenants by querying the tenants table, right now its a dirty way to build the file but its not that bad since we are storing it on kubernetes configMap, on the other side I have some concern regarding user direct input is used to create the routing rules for nginx:

```
	server {
		server_name USER-CONTROL-THIS.s3.localhost;
		location / {
			proxy_pass http://USER-PARTIALLY-CONTROL-THIS:%d;
		}
	}
```

From the security perspective we should have a mechanism to avoid malicious to expose minio internal services by defining known service names (dashboard. kubedns, etc, etc)

cluster/storage-cluster.go
```diff
@@ -187,6 +187,44 @@ func GetListOfTenantsForStorageGroup(ctx *Context, sg *StorageGroup) chan []*Sto
 	return ch
 }
 
+// GetListOfTenants returns a list of all tenants that currently exists on the cluster
```
dvaldivia: update the comment

dvaldivia (changes requested):

dvaldivia (approved):

k8s/load-balancer/Dockerfile
Alevsk: that file is not need it, im removing the whole `load-balancer` folder since it was only for testing purposes.

PR #47, opened 2019-10-21
Tenant Schema

dvaldivia:
Screen Shot 2019-10-21 at 1 57 02 PM


cesnietor (commented):
Based on yesterdays discussion, this is outdated.
Groups will not be there anymore.

dvaldivia:
Just updated it

dvaldivia:
The schema looks like this now:

Screen Shot 2019-10-22 at 4 23 09 PM

cluster/tenant-migrations/000001_tenant-initial.up.sql
```diff
@@ -0,0 +1,114 @@
+create table atenantschema.users
```
cesnietor: is `atenantschema` misspelled? or is it supposed to be like that?
Also where is this schema created?

cesnietor (approved):

cluster/tenant-migrations/000001_tenant-initial.up.sql
dvaldivia: it was never meant to be, I designed the tables on that schema, should have not been checked in

PR #61, opened 2019-10-26
Randomize Tenant's MinIO Operator Credentials

dvaldivia:
Upon tenant creation, it's operator credentials are randomized and stored on a k8s secret


Alevsk (approved):
Do we want to accept only uppercase and numbers? what about lowercase and - _ ?

Issue #34, opened 2019-10-17
Implement local user IDP


dvaldivia:
We need some basic User management features:

- [x] Register a User - return JWT token for the session
- [x] Login as a User - return JWT token for the session
- [x] Validate User Email for the provided Tenant (company)

PR #64, opened 2019-10-28
Add Login api using grpc 

cesnietor:
Solves #34 
Login api added using GRPC which returns a session hashed in a token.
Verifies if credentials exist in the defined tenant and if the user is valid and authenticated it will create a session, save it on the `provisioning` schema and return it on the grpc response.


common/utils.go
```diff
@@ -0,0 +1,43 @@
+// This file is part of MinIO Kubernetes Cloud
```
aead: What is the purpose of this package?

Actually the functionality here is not worth a separate package.
In particular it should just be:
```
b := make([]byte, size)
if _, err := io.ReadFull(rand.Reader, b); err != nil {
     // TODO: error handling
}
// now use b or encode it as base64 / hex / ...
```

cluster/setup.go
```diff
@@ -67,8 +71,37 @@ func SetupM3Namespace() {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "jwtkey",
+		},
+		Data: map[string][]byte{
+			"M3_JWT_KEY": []byte(common.GetRandString(64, "default")),
+		},
+	}
+	res, err := clientset.CoreV1().Secrets("default").Create(&secret)
+	if err != nil {
+		fmt.Println(err)
```
aead: Error handling?

cluster/setup.go
```diff
@@ -67,8 +71,37 @@ func SetupM3Namespace() {
...
+		panic(err.Error())
+	}
+	if res.Name == "" {
+		fmt.Println(err)
```
aead: Error handling?

cmd/m3/tenand-add-user.go
```diff
@@ -0,0 +1,88 @@
...
+		tenantShortName = ctx.Args().Get(0)
+	}
+	if email == "" && ctx.Args().Get(1) != "" {
+		email = ctx.Args().Get(1)
+	}
+	if password == "" && ctx.Args().Get(2) != "" {
+		password = ctx.Args().Get(2)
+	}
+	if tenantShortName == "" {
+		fmt.Println("You must provide tenant name")
```
aead: So, we print an "error message" but not return an error?

cmd/m3/tenand-add-user.go
```diff
@@ -0,0 +1,88 @@
...
+//     m3 tenant add-user tenant-1 user@acme.com user1234
+//     m3 tenant add-user --tenant tenant-1 --email user@acme.com --password user123
+func tenantAddUser(ctx *cli.Context) error {
+	tenantShortName := ctx.String("tenant")
+	email := ctx.String("email")
+	password := ctx.String("password")
+	if tenantShortName == "" && ctx.Args().Get(0) != "" {
+		tenantShortName = ctx.Args().Get(0)
+	}
+	if email == "" && ctx.Args().Get(1) != "" {
```
aead: What happens if the users does not provide 2 arguments but just 1?

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,276 @@
...
+	return config
+}
+
+// getJWTSecretKey gets jwt secret key from kubernetes secrets
+func getJWTSecretKey() ([]byte, error) {
+	config := getConfig()
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		fmt.Println(err)
```
aead: Error handling?

aead (dismissed):

aead:
This PR is quite large... Isn't it possible to split it up into 2/3 smaller ones?

cesnietor:
> This PR is quite large... Isn't it possible to split it up into 2/3 smaller ones?
yes, makes sense. Will do.

common/utils.go
cesnietor: so in this package I'm planning to put all code that is being shared among portal and cluster packages to avoid copy pasting code.

cmd/m3/tenand-add-user.go
cesnietor: fixed in pr #66

cmd/m3/tenand-add-user.go
cesnietor: If the user don't provide the arguments the next if will catch and ask the user to provide them. (Moved to #66 )

portal/api-authentication-grpc.go
cesnietor: done

common/utils.go
cesnietor: I've modified the function with the suggestion. Still since this function is being used in m3/cluster and m3/portal that's why needs to be in separate package to avoid duplicate code. What do you suggest instead of creating a new package?

cluster/setup.go
cesnietor: So far we are using panic with all setups.

common/utils.go
cesnietor: Ok I just removed lol, since m3/cluster for now will not need it anymore.

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,211 @@
...
+	Tenant   string
+	IsAdmin  bool
+	Password string
+	UUID     string
+}
+
+// Login handles the Login request by receiving the user credentials
+// and returning a hashed token.
+func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (*pb.LoginResponse, error) {
+	log.Printf("Calling Login")
```
dvaldivia: maybe we don't needs this

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,211 @@
...
+				($1,$2,$3,$4)`
+
+	// Execute Query
+	_, err = loginCtx.Tx.Exec(query, sessionID, userID, tenant.ID, time.Now())
+	if err != nil {
+		tx.Rollback()
+		res.Error = err.Error()
+		return &res, err
+	}
+	fmt.Println("sessionID: ", sessionID)
```
dvaldivia: another print xD

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,211 @@
...
+	// Search for the tenant on the database
+	tenant, err := getTenant(tenantName)
+	if err != nil {
+		err = errors.New("Tenant not found")
+		res.Error = err.Error()
+		return &res, err
+	}
+
+	// Password validation
+	user, ok := getUser(tenant.Name, email)
```
dvaldivia: this validation should be on the DB

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,211 @@
...
+	queryUser := fmt.Sprintf(`
+		SELECT 
+				t1.id, t1.email, t1.password
+			FROM 
+				tenants.%s.users t1
+			WHERE email=$1`, quoted)
+	row := loginCtx.Tx.QueryRow(queryUser, email)
+
+	// Save the resulted query on the User struct
+	err = row.Scan(&user.UUID, &userEmail, &user.Password)
```
dvaldivia: I cannot find where this `user` variable is instanciated

dvaldivia (dismissed):

portal/api-authentication-grpc.go
cesnietor: it is being created on the definition of the function 
`func getUser(tenant string, email string) (user User, ok bool)`

portal/api-authentication-grpc.go
cesnietor: removed

portal/api-authentication-grpc.go
cesnietor: Updated by lookin on the db both user email and password and returning ~"wrong credentials" error when failing doing the query.

portal/api-authentication-grpc.go
cesnietor: done

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+	Password string
+	UUID     string
+}
+
+// Login handles the Login request by receiving the user credentials
+// and returning a hashed token.
+func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (*pb.LoginResponse, error) {
+	// Create Credentials
+	// TODO: validate credentials: username->email, tenant->shortname?
+	var res pb.LoginResponse
```
aead:
```suggestion
```

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+	Tenant   string
+	Email    string
+	IsAdmin  bool
+	Password string
+	UUID     string
+}
+
+// Login handles the Login request by receiving the user credentials
+// and returning a hashed token.
+func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (*pb.LoginResponse, error) {
```
aead:
```suggestion
func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (res *pb.LoginResponse, err error) {
```

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+	// TODO: validate credentials: username->email, tenant->shortname?
+	var res pb.LoginResponse
+	tenantName := in.GetCompany()
+	email := in.GetEmail()
+	pwd := in.GetPassword()
+
+	// Search for the tenant on the database
+	tenant, err := getTenant(tenantName)
+	if err != nil {
+		err = errors.New("Tenant not found")
```
aead: Why do we overwrite the error. Shouldn't we return an `err` directly?

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+		return &res, err
+	}
+
+	// Password validation
+	// Look for the user on the database by email AND pwd,
+	// if it doesn't exist it means that the email AND password don't match, therefore wrong credentials.
+	// TODO: hash password and pass it to the getUser assuming db has hashed password also.
+	user, err := getUser(tenant.Name, email, pwd)
+	if err != nil {
+		err := errors.New("Wrong tenant, email and/or password")
```
aead: Why do we overwrite the error. Shouldn't we return an `err` directly?

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+		return &res, err
+	}
+
+	// Return session in Token Response
+	res.JwtToken = sessionID
+
+	// if no error happened to this point commit transaction
+	err = tx.Commit()
+	if err != nil {
+		res.Error = err.Error()
```
aead: Why do we overwrite the error. Shouldn't we return an `err` directly?

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+		return &res, err
+	}
+	return &res, nil
+}
+
+// getUser searches for the user in the defined tenant's database
+// and returns the User if it was found
+func getUser(tenant string, email string, password string) (user User, err error) {
+	bgCtx := context.Background()
```
aead, line 123: I would suggest:
```
defer func() {
    if err != nil {
          tx.Rollback()
    }
}
```

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+	return user, nil
+}
+
+// getTenant gets the Tenant if it exists on the m3.provisining.tenants table
+// search is done by tenant name
+func getTenant(tenantName string) (tenant cluster.Tenant, err error) {
+	bgCtx := context.Background()
+	db := cluster.GetInstance().Db
+
+	tx, err := db.BeginTx(bgCtx, nil)
```
aead, line 169: I would suggest:
```
defer func() {
    if err != nil {
          tx.Rollback()
    }
}
```

portal/api-authentication-grpc.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		tx.Rollback()
```
aead: We shouldn't use `tx` if creating a transaction failed, right?!

cesnietor: lol true, changing it on all places where I do it.

portal/api-authentication-grpc.go
cesnietor: makes sense, checking.

portal/api-authentication-grpc.go
cesnietor: Right now we haven't defined well how we are gonna handle the grpc errors, right now it will be returned on the response to specify our our own defined errors. We will update that in future versions if needed.

portal/api-authentication-grpc.go
cesnietor: modified to handle errors on response

portal/api-authentication-grpc.go
cesnietor: done

portal/api-authentication-grpc.go
cesnietor: done

portal/api-authentication-grpc.go
cesnietor: done.

portal/api-authentication-grpc.go
cesnietor: same as above. handling errors on grpc response

krisis:
LGTM. We can improve our approach to returning error from grpc server methods.

krisis (approved):

krisis (approved):

kannappanr (approved):
LGTM

PR #65, opened 2019-10-28
use a cryptographic PRNG when generating credentials

aead:
This commit replaces the implementation of generating
random access/secret keys. The previous implementation
used math/rand and a scheme from
https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go
to generate "random" access/secret keys.

This is not correct since it uses math/rand for generating
access credentials. For any kind of secret information
(cryptographic keys a.s.o) a (cryptographically) secure PRNG
must be used.

### DIFF
--- a/cluster/utils.go
+++ b/cluster/utils.go
@@ -17,36 +17,35 @@
 package cluster
 
 import (
-	"math/rand"
+	"crypto/rand"
+	"io"
 	"strings"
-	"time"
 )
 
-const letterBytes = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+// Do not use:
+// https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go
+// It relies on math/rand and therefore not on a cryptographically secure RNG => It must not be used
+// for access/secret keys.
 
-var src = rand.NewSource(time.Now().UnixNano())
+// The alphabet of random character string. Each character must be unique.
+//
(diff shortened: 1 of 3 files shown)


cluster/utils.go
```diff
@@ -17,36 +17,35 @@
...
-	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
-)
+// The alphabet of random character string. Each character must be unique.
+//
+// The RandomCharString implementation requires that: 256 / len(letters) is a natural numbers.
+// For example: 256 / 64 = 4. However, 5 > 256/62 > 4 and therefore we must not use a alphabet
+// of 62 characters.
+// The reason is that if 256 / len(letters) is not a natural number then certain characters become
+// more likely then others.
+const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_"
```
harshavardhana: we can avoid `-_` like characters to avoid bugs. We can stick to just characters

harshavardhana (changes requested):

cluster/utils.go
aead: We can only limit to e.g. upper case letters - 4 numbers or other combinations.
26 (a...z) + 10 (0...9) = 36 but `256 / 36 = 7.111...` We need an alphabet of 16, 32 or 64 chars.

dvaldivia (approved):

harshavardhana (approved):

PR #67, opened 2019-10-29
Update storage group and nginx deployment on all nodes on tenant-add

krisis:


cesnietor (dismissed):

cluster/cluster.go
```diff
@@ -759,11 +701,16 @@ func DeployNginxResolver() {
 		},
 	}
 
-	resDeployment, err := extV1beta1API(clientset).Deployments("default").Create(&deployment)
-	if err != nil {
-		panic(err.Error())
+	var resDeployment *v1beta1.Deployment
+	if shouldUpdate {
+		if resDeployment, err = extV1beta1API(clientset).Deployments("default").Update(&deployment); err != nil {
```
cesnietor: We might need to check this since it seems that the Update is not being done.
My theory is that it might be because the deployment didn't change.

krisis: @cesnietor Yes, you are right. This PR address the above issue now.

cluster/nginx.go
```diff
@@ -0,0 +1,177 @@
...
+			ch <- err
+			return
+		default:
+			shouldUpdate = true
+		}
+		DeployNginxResolver(shouldUpdate)
+	}()
+	return ch
+}
+
```
cesnietor, line 116: Can you please add comments for Docs?

cluster/nginx.go
```diff
@@ -0,0 +1,177 @@
...
+func DeleteNginxLBDeployment(clientset *kubernetes.Clientset, deploymentName, appLabel string) <-chan struct{} {
+	// Setup shared pods informer on default namespace to detect the
+	// completion of the deletion of pods of nginx-resolver
+	// deployment
+	doneCh := make(chan struct{})
+	go func() {
+		// Setup shared pods informer on default namespace to detect the
+		// completion of the deletion of pods of nginx-resolver
+		// deployment
```
cesnietor: duplicate comments

cluster/nginx.go
```diff
@@ -0,0 +1,177 @@
...
+		default:
+			shouldUpdate = true
+		}
+		DeployNginxResolver(shouldUpdate)
+	}()
+	return ch
+}
+
+func DeleteNginxLBDeployment(clientset *kubernetes.Clientset, deploymentName, appLabel string) <-chan struct{} {
+	// Setup shared pods informer on default namespace to detect the
```
cesnietor: Is now a deployment informer?

cluster/cluster.go
```diff
@@ -199,12 +198,13 @@ type TenantConfiguration struct {
 func CreateTenantSecrets(tenant *Tenant, tenantConfig *TenantConfiguration) error {
```
cesnietor, line 198: should this function be also on `minio-tenant.go` file and all related tenant functions?

cluster/nginx.go
krisis: Yes. I have added the comments.

cesnietor (approved):

PR #68, opened 2019-11-01
Create user ui credentials

dvaldivia:
What's going on on this PR?

- The functionality to create the UI Credentials on `add user` was implemented.
  - When the UI Credential is generated it is committed to the Tenant's MinIO
  - The new `MinIO User` or access key, is granted `readwrite` policy
  - The secret key is stored on k8s
- Command `./m3 tenant add-user` was renamed to `./m3 tenant user add` 

To achieve these simple things one thing had to be fixed, we have a main schema (m3.provisioning) and each tenant has it's own schema, we were opening a `cluster.Context` based on what we planned to query/insert, however adding a credential represents the first mix of the two, therefore I made it so Context houses both Tenant DB Transactions and Main DB Transactions, moreover this transactions are not started until they are requested via `ctx.TenantTx()` or `ctx.MainTx()`, additionally, for scenarios where we only need to read from either DB and not insert, we don't need to open a transaction for the database, here the tenant DB can be reached using `ctx.TenantDB()` and the main db can be acccessed via `cluster.GetInstance().Db`

I had to adjust all the code that relies on `cluster.Context` and align it with the new db paradigm.

It's important to note that the `ctx.TenantDB()` and `ctx.TenantTx()` objects are already configured to use the proper database and schema, so it's not needed to include that in the query, it was error prone to have to put it in on every query/function. On subsequent transactions `m3.provisioning` will be removed from Main DB queries.

Also we need to honor the `ctx.WhoAmI` and always insert it to the `created_by` field on all the tenant tables.

### DIFF
--- a/cluster/context.go
+++ b/cluster/context.go
@@ -35,7 +35,7 @@ type Context struct {
 	WhoAmI string
 }
 
-// MainTx returns a transaction agains the Main DB, if none has been started, it starts one
+// MainTx returns a transaction against the Main DB, if none has been started, it starts one
 func (c *Context) MainTx() (*sql.Tx, error) {
 	if c.mainTx == nil {
 		db := GetInstance().Db
@@ -57,7 +57,7 @@ func (c *Context) TenantDB() *sql.DB {
 	return c.tenantDB
 }
 
-// TenantTx returns a transaction agains the Tenant DB, if none has been started, it starts one
+// TenantTx returns a transaction against the Tenant DB, if none has been started, it starts one
 func (c *Context) TenantTx() (*sql.Tx, error) {
 	if c.mainTx == nil {
 		db := c.TenantDB()
--- a/cluster/context.go
+++ b/cluster/context.go
@@ -27,11 +27,13 @@ import (
 type Context struct {
 	*sql.Tx
 	Main context.Context
+	// a user identifier of who is starting the context
+	WhoAmI string
 }
 
 // Creates a new `Context` given an initial transaction and `context.Context`
 // to control timeouts and cancellations.
 func NewContext(ctx context.Context, tx *sql.Tx) *Context {
-	c := &Context{Tx: tx, Main: ctx}
+	c := &Context{Tx: tx, Main: ctx, WhoAmI: "command line"}
 	return c
 }
--- a/cluster/context.go
+++ b/cluster/context.go
@@ -35,6 +35,7 @@ type Context struct {
 	WhoAmI string
 }
 
+// MainTx returns a transaction agains the Main DB, if none has been started, it starts one
 func (c *Context) MainTx() (*sql.Tx, error) {
 	if c.mainTx == nil {
 		db := GetInstance().Db
@@ -47,6 +48,7 @@ func (c *Context) MainTx() (*sql.Tx, error) {
 	return c.mainTx, nil
 }
 
+// TenantDB returns a configured DB connection for the Tenant DB
 func (c *Context) TenantDB() *sql.DB {
 	if c.tenantDB == nil {
 		db := GetInstance().GetTenantDB(c.TenantShortName)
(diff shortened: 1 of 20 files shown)


cluster/tenant-users.go
```diff
@@ -0,0 +1,127 @@
...
+	err = tx.Commit()
+	if err != nil {
+		return nil
+	}
+	return nil
+}
+
+// GetUserWithPwd searches for the user in the defined tenant's database
+// and returns the User if it was found
+func GetUserWithPwd(ctx *Context, tenant string, email string, password string) (user User, err error) {
```
cesnietor: nice

cluster/context.go
```diff
@@ -25,13 +25,102 @@ import (
...
+		if err != nil {
+			return err
+		}
+		// restart the txn
+		c.mainTx = nil
+	}
+	return nil
+}
+
+func (c *Context) Rollback() error {
```
cesnietor, line 96: if we want to just rollback the TenanTx this will rollback both mainTx and Tenantx or are they mutually exclusive?

cluster/tenant-users.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	// Create this user's credentials so he can interact with it's own buckets/data
+	err = createUserCredentials(ctx, tenantShortName, userID)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	// if no error happened to this point commit transaction
+	err = tx.Commit()
```
cesnietor: Is this `tx.Commit()` also committing createUserCredentials transactions? or where is that being commited?

cluster/tenant-users.go
dvaldivia: +1

cluster/credentials.go
```diff
@@ -0,0 +1,154 @@
...
+// tenant DB
+func createUserCredentials(ctx *Context, tenantShortName string, userdID uuid.UUID) error {
+
+	userUICredentials := UserUICredentials{
+		AccessKey: RandomCharString(16),
+		SecretKey: RandomCharString(32)}
+
+	// Attempt to store in k8s, if it works, store in DB
+
+	err := storeUserUICredentialsSecret(tenantShortName, &userdID, &userUICredentials)
```
cesnietor, line 46: if we fail after this part will this secret still going to be there?

cluster/context.go
dvaldivia: you can get each transaction individually and roll it back

cluster/tenant-users.go
dvaldivia: nice catch, will update

cluster/credentials.go
dvaldivia: yes, until we come up with a rollback strategy for kubernetes

cluster/tenant-users.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Get user from tenants database
+	queryUser := `
+		SELECT 
+				t1.id, t1.email, t1.password, t1.is_admin
+			FROM 
+				users t1
+			WHERE email=$1 AND password=$2`
+
+	row := ctx.TenantDB().QueryRow(queryUser, email, hashedPassword)
```
Alevsk: How does this select works? I understand **hashedPassword** function uses the **bcrypt** one-way hashing function that generates a random salt every time, if this works the same way as in other languages using the hashed password like this in the query would never retrieve any result, instead we need to use a **bcrypt compare** function, ie: https://coderrocketfuel.com/article/using-bcrypt-to-hash-and-check-passwords-in-node-js#check-a-user-entered-password

Alevsk (changes requested):

cluster/tenant-users.go
dvaldivia: this way the question is sent to the DB  and we don't have to pull any data from the DB is the password is wrong

Alevsk: this: https://stackoverflow.com/questions/23259586/bcrypt-password-hashing-in-golang-compatible-with-node-js#answer-23259804

Alevsk (dismissed):
Looks good.

Alevsk (approved):
Passed all tests

PR #69, opened 2019-11-03
Delete tenant command

dvaldivia:
Introduces `./m3 tenant delete TENANNAME --confirm` command to remove a tenant from the m3 cluster.

The command starts by removing the tenant from the tenant table and the starts removing all the tenant configurations (secrets, services), removes the tenant from the router and finally redeploys the storage group it resides on. 

`ReDeployStorageGroup` had to be updated as removing tenants and adding new ones was causing a problem where duplicate or ghost volumes were being kept

### DIFF
--- a/cluster/minio-tenant.go
+++ b/cluster/minio-tenant.go
@@ -29,8 +29,9 @@ func mkTenantMinioContainer(sgTenant *StorageGroupTenant, hostNum string) (v1.Co
 	envName := fmt.Sprintf("%s-env", sgTenant.ShortName)
 	volumeMounts := []v1.VolumeMount{}
 	tenantContainer := v1.Container{
-		Name:            fmt.Sprintf("%s-minio-%s", sgTenant.Tenant.ShortName, hostNum),
-		Image:           "minio/minio:latest",
+		Name:  fmt.Sprintf("%s-minio-%s", sgTenant.Tenant.ShortName, hostNum),
+		Image: "minio/minio:RELEASE.2019-10-12T01-39-57Z",
+		//Image:           "minio/minio:latest",
 		ImagePullPolicy: "IfNotPresent",
 		Args: []string{
 			"server",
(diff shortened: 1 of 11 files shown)


cesnietor:
in the current master this error occurs when trying to add a user.
Is this being solved here?
```
./m3 tenant user add acme cesnietor@acme.com cesnietor
Error adding user: Put http://acme-sg-1:9001/minio/admin/v1/add-user?accessKey=VFKCTFC0U5S52TKG: dial tcp: lookup acme-sg-1: no such host
```

dvaldivia:
I have a fix for that already, found several bugs, but I have not pushed them yet, will do so tonight

dvaldivia:
@cesnietor i just pushed the fix for that to this PR

cluster/minio-tenant.go
```diff
@@ -29,8 +29,9 @@ func mkTenantMinioContainer(sgTenant *StorageGroupTenant, hostNum string) (v1.Co
 	envName := fmt.Sprintf("%s-env", sgTenant.ShortName)
 	volumeMounts := []v1.VolumeMount{}
 	tenantContainer := v1.Container{
-		Name:            fmt.Sprintf("%s-minio-%s", sgTenant.Tenant.ShortName, hostNum),
-		Image:           "minio/minio:latest",
+		Name:  fmt.Sprintf("%s-minio-%s", sgTenant.Tenant.ShortName, hostNum),
+		Image: "minio/minio:RELEASE.2019-10-12T01-39-57Z",
```
Alevsk, line 33: Do we want to keep using this specific version of minio?

Alevsk (approved):

cluster/minio-tenant.go
dvaldivia: because after this release, next version has breaking changes, so we want to make sure we upgrade when we are ready and we don't start seeing random results, we need to update madmin for example to be able to work with the next version of MinIO (i.e. minio:edge)

harshavardhana: there are breaking changes that's why it's better to stick to a version i.e tested and stable @Alevsk @dvaldivia - m3 can migrate to future releases as it gets tested etc and comes under m3's own release cycles.

Perhaps we can make this optional in future too, where users can choose their right version of MinIO

PR #71, opened 2019-11-04
Router Refresh command and Fixes

dvaldivia:
Introduces command to refresh the cluster router, this is useful when the cluster is in a bad state and you want to force an nginx configuration refresh.

This also introduces a default server on the nginx config so we return 404 if a bad address is accessed.


Alevsk (approved):
looks good

PR #73, opened 2019-11-05
portal: Add MakeBucket grpc handler

krisis:
Add portal-client-grpc-example.go as an instructive example for grpc client

### DIFF
--- a/portal/api-authentication-grpc.go
+++ b/portal/api-authentication-grpc.go
@@ -19,6 +19,8 @@ package portal
 import (
 	"context"
 
+	"database/sql"
+
 	"golang.org/x/crypto/bcrypt"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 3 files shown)


portal/api-authentication-grpc.go
```diff
@@ -19,12 +19,94 @@ package portal
...
+		return nil, status.New(codes.Unauthenticated, "SessionId not found").Err()
+	default:
+		sessionId = sIds[0]
+	}
+
+	// With validating sessionId behind us, we query the tenant MinIO
+	// service corresponding to the logged-in user to make the bucket
+
+	// Prepare DB instance
+	db := cluster.GetInstance().Db
```
dvaldivia, line 53: Can you check `cluster.MakeBucket()` I think that function already does most of what is implemented here

portal/api-authentication-grpc.go
```diff
@@ -19,12 +19,94 @@ package portal
...
+		tenantConf.AccessKey,
+		tenantConf.SecretKey,
+		false)
+	if err != nil {
+		return nil, status.New(codes.Internal, "Failed to connect to object storage").Err()
+	}
+
+	// Create bucket in the tenant's MinIO
+	// TODO: Need to decide if we are going to support regions
+	bucket := in.GetName()
```
dvaldivia, line 97: bucket name needs validation, or do you want MinIO to error out and use that as validation?

dvaldivia (changes requested):

portal/api-authentication-grpc.go
krisis: Yes, let me check that. On a related note, minio-go make bucket errors out because it is not able to resolve tenant address on the local development setup.

portal/api-authentication-grpc.go
krisis: I would rely on Minio's validation unless m3 has more restrictions on bucket names.

dvaldivia: that error is because you need to port-forward the service to your local and add it to `/etc/hosts` alternatively you can use `kubefwd` so you local can reach the desired MinIO

portal/api-authentication-grpc.go
dvaldivia: (put the other comment on the wrong thread) that is because you need to port-forward the MinIO service to your local and add it to your `/etc/hosts`, alternatively you can use `kubefwd`, once M3 is running inside k8s it will have direct access to the services

cesnietor: added a pr #75 on how to use kubefwd while developing

krisis: > Can you check `cluster.MakeBucket()` I think that function already does most of what is implemented here

done

cesnietor (dismissed):
LGTM

dvaldivia (dismissed):

portal/api-authentication-grpc.go
```diff
@@ -91,36 +165,17 @@ func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (res *pb.LoginR
...
+	var (
+		err          error
+		appCtx       *cluster.Context
+		sessionRowID string
+	)
+	if sessionRowID, err = validateSessionID(ctx); err != nil {
+		return &pb.Empty{}, nil
 	}
 
-	// If session exists on db and is valid, then proceed to update it's status to invalid
```
cesnietor, line 121: my comment :(

cesnietor (approved):
good

PR #80, opened 2019-11-06
Add DeleteBucket grpc handler

krisis:


cluster/tenants.go
```diff
@@ -432,6 +432,42 @@ func ListBuckets(tenantShortname string) ([]string, error) {
...
+		return err
+	}
+
+	var (
+		tenantConf *TenantConfiguration
+		sgt        *StorageGroupTenantResult
+	)
+	sgt, tenantConf, err = getMinioTenantInfo(ctx, tenantShortname)
+	if err != nil {
+		ctx.Rollback()
```
cesnietor: is there any transaction to rollback?

cluster/tenants.go
```diff
@@ -314,18 +314,16 @@ func MigrateTenantDB(tenantName string) chan error {
 	return ch
 }
 
-func getMinioTenantInfo(ctx *Context, tenantShortname string) (*StorageGroupTenantResult, *TenantConfiguration, error) {
+func getMinioTenantInfo(tenantShortname string) (*StorageGroupTenantResult, *TenantConfiguration, error) {
```
dvaldivia: maybe we should rethink the name if it returns both the storage group the tenant is located and the tenant general purpose configuration, I don't think it's actually the `minio info` since we don't call into MinIO, but instead it's how to reach it and how to access it

dvaldivia (dismissed):

cluster/tenants.go
krisis: @dvaldivia I agree. I have renamed this helper function as `newTenantMinioClient` which returns `*minio.Client`. PTAL.

cesnietor (approved):

PR #81, opened 2019-11-06
Service Accounts Add/List Commands

dvaldivia:


cluster/tenant-service-accounts.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+type ServiceAccount struct {
+	ID          uuid.UUID
+	Name        string
+	Description *string
+	AccessKey   string
+}
+
+// AddUser adds a new user to the tenant's database
```
cesnietor: please update this comments

cluster/tenant-service-accounts.go
```diff
@@ -0,0 +1,124 @@
...
+				($1,$2,$3,$4)`
+	tx, err := ctx.TenantTx()
+	if err != nil {
+		return nil, err
+	}
+	stmt, err := tx.Prepare(query)
+	if err != nil {
+		ctx.Rollback()
+		return nil, err
+	}
```
cesnietor, line 51: can we implement her using defer?
like:
```
defer func() {
		if err != nil {
			ctx.Rollback()
			return
		}
		// if no error happened to this point commit transaction
		err = ctx.Commit()
	}()
```

cluster/tenant-service-accounts.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// if no error happened to this point commit transaction
+	err = ctx.Commit()
+	if err != nil {
+		return nil, err
+	}
+	return sa, nil
+}
+
+// GetUsersForTenant returns a page of users for the provided tenant
```
cesnietor: please update comments `GetServiceAccountsForTenant `

cluster/tenant-service-accounts.go
```diff
@@ -0,0 +1,124 @@
...
+func GetServiceAccountsForTenant(ctx *Context, offset int, limit int) ([]*ServiceAccount, error) {
+	if offset < 0 || limit < 0 {
+		return nil, errors.New("invalid offset/limit")
+	}
+
+	// Get user from tenants database
```
cesnietor: same comments I think copy pasted

cmd/m3/tenant-service-account-list.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+		cli.IntFlag{
+			Name:  "limit",
+			Value: 20,
+			Usage: "Results per page, maximum 100.",
+		},
+	},
+}
+
+// tenantUserList lists the users on the tenant, supports pagination via offset and limit
```
cesnietor: comments update
```suggestion
// tenantServiceAccountList lists the service accounts on the tenant, supports pagination via offset and limit
```

cmd/m3/tenant-service-account-add.go
```diff
@@ -0,0 +1,99 @@
...
+			Value: "",
+			Usage: "optional description of the service account purpose",
+		},
+	},
+}
+
+// tenantAddUser Command to add a user to the tenant's database.
+// sample usage:
+//     m3 tenant add-user tenant-1 user@acme.com user1234
+//     m3 tenant add-user --tenant tenant-1 --email user@acme.com --password user123
```
cesnietor: this needs to be updated

cmd/m3/tenant-service-account-add.go
```diff
@@ -0,0 +1,99 @@
...
+package main
+
+import (
+	"fmt"
+
+	"github.com/minio/cli"
+	"github.com/minio/m3/cluster"
+)
+
+// Adds a user to the tenant's database
```
cesnietor: basically all comments please check

cluster/tenant-service-accounts.go
dvaldivia: seems like I would have to return named parameters, but since the service account is generated by an inner function I cannot use this style of deferring

cesnietor (dismissed):

cesnietor (approved):

PR #82, opened 2019-11-06
Introduce Sys Fields

dvaldivia:


cmd/m3/cluster-sc-sg-add.go
```diff
@@ -45,17 +45,25 @@ func addStorageGroup(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
 	// create a new storage group in the DB
-	storageGroupResult := <-cluster.AddStorageGroup(name)
+	storageGroupResult := <-cluster.AddStorageGroup(appCtx, name)
 	if storageGroupResult.Error != nil {
 		fmt.Println(storageGroupResult.Error)
+		appCtx.Rollback()
```
cesnietor, line 57: [1] here there will be nothing to rollback if addStorageGroup already rolled back

cluster/storage-cluster.go
```diff
@@ -38,35 +38,41 @@ type StorageGroupResult struct {
...
 		}
 		defer stmt.Close()
 
-		err = stmt.QueryRow(sgID, sgName).Scan(&sgNum)
+		err = stmt.QueryRow(sgID, sgName, ctx.WhoAmI).Scan(&sgNum)
 		if err != nil {
 			ch <- StorageGroupResult{
 				Error: err,
 			}
+			tx.Rollback()
```
cesnietor: this rollback collides with the one made on `cmd/m3/cluster-sc-sg-add.go ` see comment tagged as [1]

cmd/m3/cluster-sc-sg-add.go
dvaldivia: It's harmless if things get rolled back twice the `ctx.Rollback()` will handle being tried to rollback a non existing transaction (as it gets closed on the previous rollback)

dvaldivia: also if the internal call doesn't rollback the invoking function may decide to rollback, so it's fine to have both

cesnietor (approved):

PR #83, opened 2019-11-06
Reorganize Bucket Commands

dvaldivia:
Renames `./m3 tenant mb` to `./m3 tenant bucket add`


cesnietor (changes requested):
Can you please also update the Readme related command

cesnietor (approved):

PR #84, opened 2019-11-06
mkube admins to control access to the backend

dvaldivia:
This resource will be the base to control access to the mkube backend.

An `admin` will be the one that can issue `./m3` commands, he will need to authenticate himself signing his commands with his secret, initially we can use HMAC for that, and later on move to issue certificates or something stronger.

Upon setup of `mkube` an access/secret key is generated (these keys have nothing to do with MinIO access/secret keys) and stored on kubernetes.

This PR contains the table, and the command to add a new admin.

Subsequent PRs will introduce the authentication mechanism, where for an `m3` command to run, the provided credentials must be present either as an environment variable or in a file.

### DIFF
--- a/cluster/migrations/000001_m3-initial.up.sql
+++ b/cluster/migrations/000001_m3-initial.up.sql
@@ -9,7 +9,7 @@ create table provisioning.admins
     email varchar(256) not null,
     access_key         varchar(256)                           not null,
     sys_created_by varchar(256) not null,
-    sys_date_created timestamptz default now() not null
+    sys_created_date timestamptz default now() not null
 );
 
 create unique index admins_email_uindex
--- a/cluster/migrations/000001_m3-initial.up.sql
+++ b/cluster/migrations/000001_m3-initial.up.sql
@@ -1,5 +1,23 @@
 create schema provisioning;
 
+create table provisioning.admins
+(
+    id uuid
+        constraint admins_pk
+            primary key,
+    name varchar(256),
+    email varchar(256) not null,
+    access_key         varchar(256)                           not null,
+    sys_created_by varchar(256) not null,
+    sys_date_created timestamptz default now() not null
+);
+
+create unique index admins_email_uindex
+    on provisioning.admins (email);
+
+create index admins_access_key_uindex
+    on provisioning.admins (access_key);
+
 create table provisioning.tenants
 (
     id         uuid         not null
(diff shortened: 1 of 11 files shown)


cmd/m3/admin-add.go
```diff
@@ -0,0 +1,82 @@
...
+			Name:  "email",
+			Value: "",
+			Usage: "user email",
+		},
+	},
+}
+
+// adminAdd is a command to add a cluster admin.
+// sample usage:
+//     m3 admin add "User Name" user@acme.com
```
cesnietor, line 47: can you please add this new feature to the readme?

cluster/migrations/000001_m3-initial.up.sql
```diff
@@ -1,5 +1,23 @@
...
+create table provisioning.admins
+(
+    id uuid
+        constraint admins_pk
+            primary key,
+    name varchar(256),
+    email varchar(256) not null,
+    access_key         varchar(256)                           not null,
+    sys_created_by varchar(256) not null,
+    sys_date_created timestamptz default now() not null
```
cesnietor: all other tables have `sys_create_date`
```suggestion
    sys_created_date timestamptz default now() not null
```

cmd/m3/setup.go
```diff
@@ -32,6 +34,24 @@ var setupCmd = cli.Command{
...
+		fmt.Println("An admin name is needed")
+		return errMissingArguments
+	}
+
+	if email == "" {
+		fmt.Println("An admin email is needed")
+		return errMissingArguments
+	}
+	cluster.SetupM3(name, email)
 	return nil
```
cesnietor:
```suggestion
	return cluster.SetupM3(name, email)
```

cesnietor (changes requested):

cmd/m3/admin-add.go
dvaldivia: done

cmd/m3/setup.go
```diff
@@ -32,6 +34,24 @@ var setupCmd = cli.Command{
...
+	if name == "" {
+		fmt.Println("An admin name is needed")
+		return errMissingArguments
+	}
+
+	if email == "" {
+		fmt.Println("An admin email is needed")
+		return errMissingArguments
+	}
+	cluster.SetupM3(name, email)
```
cesnietor: oh lol it's being called twice
```suggestion
```

dvaldivia: lol, I saw your suggestion and I only thought "yeah retuning the error makes sense" hahaha

cesnietor (approved):

PR #86, opened 2019-11-06
Add list users grpc handler and proto modified to handle limit

cesnietor:


portal/api-user-grpc.go
```diff
@@ -51,3 +52,36 @@ func (s *server) AddUser(ctx context.Context, in *pb.AddUserRequest) (*pb.User,
...
+	if err != nil {
+		return nil, err
+	}
+	// Get list of users set maximum 25 per page
+	users, err := cluster.GetUsersForTenant(appCtx, reqOffset, reqLimit)
+	if err != nil {
+		return nil, status.New(codes.Internal, "Error getting Users").Err()
+	}
+	// Get total of users
+	total, err := cluster.GetTotalNumberOfUsers(appCtx)
```
krisis, line 78: We could use a different pagination technique later where we can avoid getting the total number of users. For now this works.

krisis (approved):

PR #93, opened 2019-11-07
Validate session expiration timestamp on GRPC calls

cesnietor:
Invalidate session it if it has expired

### DIFF
--- a/portal/utils.go
+++ b/portal/utils.go
@@ -28,8 +28,7 @@ import (
 )
 
 const (
-	sessionValid   = "valid"
-	sessionInvalid = "invalid"
+	sessionValid = "valid"
 )
 
 // UTCNow - returns current UTC time.
--- a/portal/utils.go
+++ b/portal/utils.go
@@ -27,6 +27,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	sessionValid   = "valid"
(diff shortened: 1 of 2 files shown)


kaankabalak (dismissed):

portal/utils.go
```diff
@@ -56,15 +61,46 @@ func getSessionRowIDAndTenantName(ctx context.Context) (string, string, error) {
...
+		getSessionIDQ := `SELECT s.id
+                           FROM m3.provisioning.sessions as s 
+                           WHERE s.id=$1 AND s.status=$2 AND NOW() >= s.expires_at`
+
+		tenantRow := db.QueryRow(getSessionIDQ, sessionID, sessionValid)
+		err = tenantRow.Scan(&sessionRowID)
+		// If session id exists, change the status to invalid
+		if err == nil && sessionRowID != "" {
+			setSessionStatusQ :=
+				`UPDATE m3.provisioning.sessions
```
krisis: We shouldn't update the session status when we are fetching session related information. The session status indicates whether or not a user is logged-in while the expires_at column signifies until what time it is valid. So, there is no need for an expired session to be marked invalid.

cesnietor: makes sense, let's keep both columns separate and not linked.

dvaldivia (approved):

PR #94, opened 2019-11-08
Move setup of db and creation of admin to m3-setup-db

krisis:
Currently, `m3 setup "Admin name" "admin@email.com"` doesn't work unless one has a local postgres server running. This PR makes `./m3 setup ...` commands work with the postgres running inside the kind based kubernetes cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,26 +83,34 @@ make m3
 - Run `m3 setup` on the local kubernetes
 
 ```
-./m3 setup "Admin Name" admin@email.com
+./m3 setup
 ```
 
-This will setup `m3` and create the first admin account **Note down the admin access/secret key**
+- Make postgres reachable from host OS
 
-You may see the following error message
+```
+kubectl port-forward -n m3 svc/postgres 5432
+```
+
+- To setup db
+
+```
+./m3 setup db
+```
+
+You may see the following error message since at times the postgres container is not running yet.
 ```
 Running Migrations
 2019/10/17 12:02:50 error connecting to database or reading migrations
 2019/10/17 12:02:50 dial tcp 127.0.0.1:5432: connect: connection refused
 ```
 
-This is benign can be fixed with the following steps
+If you do, try again in some time.
 
+- To create the first admin account for `m3`
+**Note down the admin access/secret key**
 ```
-kubectl port-forward -n m3 svc/postgres 5432
-```
-
-```
-./m3 setup db
+./m3 admin add "Admin" "admin@email.com"
 ```
 
 ## Creating a new Storage Group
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -67,28 +67,31 @@ make m3
...
 ```
 
-This will setup `m3` and create the first admin account **Note down the admin access/secret key**
+- Make postgres reachable from host OS
 
-You may see the following error message
 ```
-Running Migrations
-2019/10/17 12:02:50 error connecting to database or reading migrations
-2019/10/17 12:02:50 dial tcp 127.0.0.1:5432: connect: connection refused
```
cesnietor, line 95: Since `SetupM3` function is still doing `setupPostgres()` this  message should still be after `./m3 setup` since otherwise you won't know if you need to do the port-forward or not.

cesnietor: oh forget about what I said, I realized `RunMigrations ` was removed from `./m3 setup`

cesnietor (dismissed):

dvaldivia (changes requested):
The idea of “m3 setup db” is to have a command to run database migrations exclusively.

I think adding the first admin on setup is not a problem since the real setup will run inside kubernetes and Postgres will be reachable, if this is a problem during development let’s introduce a new command that finishes the setup but let’s keep a command only for database migrations.

krisis:
@dvaldivia OK. IIUC, we should add first admin as part of `mc setup db`.  Let me add, `mc admin add --name "Administrator" --email "admin@email.com"` to add the first admin. In local development, we should call this after `mc setup db`.

dvaldivia (dismissed):

README.md
```diff
@@ -83,26 +83,35 @@ make m3
...
-This will setup `m3` and create the first admin account **Note down the admin access/secret key**
+- Make postgres reachable from host OS
 
-You may see the following error message
+```
+kubectl port-forward -n m3 svc/postgres 5432
+```
+
+- To setup db
+**Note down the admin access/secret key**
```
cesnietor:
```suggestion
```

README.md
```diff
@@ -83,26 +83,35 @@ make m3
...
 
-```
-kubectl port-forward -n m3 svc/postgres 5432
-```
+- To create the first admin account for `m3`
 
 ```
-./m3 setup db
+./m3 admin add "Admin" "admin@email.com"
 ```
```
cesnietor, line 115:
```suggestion
**Note down the admin access/secret key**
```

cesnietor (changes requested):

cesnietor (approved):
LGTM

PR #95, opened 2019-11-08
Add ChangeBucketAccess grpc handler

krisis:


cesnietor (dismissed):
We might need to re do the stubs  since #96 will be merged

cesnietor (approved):

PR #97, opened 2019-11-11
Invite email

dvaldivia:
Introduces mailing capabilities to invite users to setup their password, right now, the flow to generate a token and mails the newly added user.

There two ways to get a user invited:

Upon tenant creation via `./m3 tenant add --name tenant --short_name tenant --admin_name "Daniel Valdivia" --admin_email daniel@blabla.io`

Upon adding a user `./m3 tenant user add --name "Daniel Valdivia" --email daniel@blabla.io --invite`

Makes setting a password for a new user, optional via `--password` flag


Dockerfile
```diff
@@ -8,14 +8,14 @@ RUN go mod download
 
 ADD . /go/src/github.com/minio/m3/
 WORKDIR /go/src/github.com/minio/m3/
-WORKDIR /
 
-RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-w -s" -a -installsuffix cgo -o m3 ./go/src/github.com/minio/m3/
+
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-w -s" -a -installsuffix cgo -o m3 ./cmd/m3
```
harshavardhana: installsuffix is not useful https://www.reddit.com/r/golang/comments/a8uhnd/is_installsuffix_cgo_needed_with_cgo_enabled0/ - these are old stuff which is not needed.

cluster/tenant-users.go
```diff
@@ -211,3 +216,84 @@ func GetTotalNumberOfUsers(ctx *Context) (int, error) {
...
+	query := `UPDATE users SET password=$1 WHERE id=$2`
+	tx, err := ctx.TenantTx()
+	if err != nil {
+		return err
+	}
+	stmt, err := tx.Prepare(query)
+	if err != nil {
+		ctx.Rollback()
+		return err
+	}
```
cesnietor: [1]Since statements make sense for repeated executions, it makes no sense to use it here because we are also not using stmt.Exec()
```
A prepared statement is a SQL statement with parameter placeholders which is sent to the database server and prepared for repeated execution.
```
https://mindbowser.com/golang-go-database-sql/

cluster/url-tokens.go
```diff
@@ -0,0 +1,121 @@
...
+	tx, err := ctx.MainTx()
+	if err != nil {
+		return nil, err
+	}
+	stmt, err := tx.Prepare(query)
+	if err != nil {
+		ctx.Rollback()
+		return nil, err
+	}
+	defer stmt.Close()
```
cesnietor: same as [1]

cluster/url-tokens.go
```diff
@@ -0,0 +1,121 @@
...
+	tx, err := ctx.MainTx()
+	if err != nil {
+		return err
+	}
+	stmt, err := tx.Prepare(query)
+	if err != nil {
+		ctx.Rollback()
+		return err
+	}
+	defer stmt.Close()
```
cesnietor: same as [1]

cesnietor (changes requested):

cesnietor (changes requested):
Since we can create a User with an empty password `""` you can actually login with an empty password, we shouldn't allow that.

cluster/tenant-users.go
```diff
@@ -123,14 +125,8 @@ func SetUserEnabled(tenantShortName string, userID string, status bool) error {
...
 	}
-	stmt, err := tx.Prepare(query)
-	if err != nil {
-		ctx.Rollback()
-		return err
-	}
-	defer stmt.Close()
 	// Execute query
-	_, err = stmt.Exec(status, userID)
+	_, err = tx.Exec(query, userID)
```
cesnietor:
```suggestion
	_, err = tx.Exec(query, status, userID)
```

cesnietor (approved):

PR #98, opened 2019-11-12
Spawning new nginx and deleting previous deployment versions

Alevsk:
When creating new tenants we update the MapConfig stored in Kubernetes, then create a new version of Nginx deployment that will take that new configuration, we update the nginx-resolver service to start routing traffic to the new pods (owned by the new deployment) and finally we kill all outdated Nginx deployments/pods.


dvaldivia (approved):

cluster/nginx.go
```diff
@@ -143,15 +131,26 @@ func DeleteNginxLBDeployment(clientset *kubernetes.Clientset, deploymentName, ap
...
+		}
+		for _, deployment := range deployments.Items {
+			if deployment.Name != deploymentName {
+				// This delete option is to make sure that the pods belonging to
+				// the nginx-resolver deployment are 'collected' immediately.
+				fgPropagation := metav1.DeletePropagationForeground
+				fgDeleteOption := metav1.DeleteOptions{
+					PropagationPolicy: &fgPropagation,
+				}
+				err := extV1beta1API(clientset).Deployments("default").Delete(deployment.Name, &fgDeleteOption)
```
cesnietor, line 149: I'm wondering if only the first element of the `for` will be considered since once the Delete of the first Item finishes it will close the channel on the Informer (line 127), therefore continuing with the wait code on `DeployNginxResolver` line 193. is that correct?

PR #101, opened 2019-11-12
using informers to track nginx-resolver deployment and service creation and updates

Alevsk:
Now we wait until the nginx-resolver deployment is created, after that we update the nginx-resolver service and finally we remove the old nginx-resolver deployments that are outdated (we dont need to wait for that since all the traffic is already redirected to the new nodes)


cluster/nginx.go
```diff
@@ -140,18 +117,71 @@ func DeleteNginxLBDeployments(clientset *kubernetes.Clientset, deploymentName st
 		}
 		for _, deployment := range deployments.Items {
 			if deployment.Name != deploymentName {
-				// This delete option is to make sure that the pods belonging to
-				// the nginx-resolver deployment are 'collected' immediately.
 				fgPropagation := metav1.DeletePropagationForeground
 				fgDeleteOption := metav1.DeleteOptions{
 					PropagationPolicy: &fgPropagation,
 				}
```
krisis: There is no need to force the deletion of the pods (and containers) belonging to the 'older' nginx-resolver deployment immediately, since we already have a new one in its place. We can use the default delete option of 'lazily' deleting the pods.

cluster/nginx.go
```diff
@@ -140,18 +117,71 @@ func DeleteNginxLBDeployments(clientset *kubernetes.Clientset, deploymentName st
...
+	return doneCh
+}
+
+func CreateNginxResolverDeployment(clientset *kubernetes.Clientset, deploymentName string) <-chan struct{} {
+	doneCh := make(chan struct{})
+	go func() {
+		factory := informers.NewSharedInformerFactory(clientset, 0)
+		deploymentInformer := factory.Extensions().V1beta1().Deployments().Informer()
+		deploymentInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
```
krisis: The add event doesn't mean that the pods of the deployment are in "Running" status. It is fool proof to check if the status of the 'newObj` is "Running' in UpdateFunc handler.

cluster/nginx.go
```diff
@@ -140,18 +117,71 @@ func DeleteNginxLBDeployments(clientset *kubernetes.Clientset, deploymentName st
...
+
+func UpdateNginxResolverService(clientset *kubernetes.Clientset) <-chan struct{} {
+	doneCh := make(chan struct{})
+	go func() {
+		factory := informers.NewSharedInformerFactory(clientset, 0)
+		serviceInformer := factory.Core().V1().Services().Informer()
+		serviceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, obj interface{}) {
+				service := obj.(*v1.Service)
+				if service.GetLabels()["name"] == "nginx-resolver" {
```
krisis: To confirm that the UpdateFunc is seeing the effect of the update from this function, we should check that the updated selector is same as what this function updated it to.

cluster/nginx.go
```diff
@@ -167,35 +197,18 @@ func DeployNginxResolver(shouldUpdate bool) error {
...
-		if _, err = extV1beta1API(clientset).Deployments("default").Create(&nginxLBDeployment); err != nil {
-			return err
-		}
-	}
+	nginxResolverVersion = fmt.Sprintf(`%s-%s`, nginxLBName, strings.ToLower(RandomCharString(6)))
+	waitCreateCh := CreateNginxResolverDeployment(clientset, nginxResolverVersion)
+	<-waitCreateCh
+	waitUpdateCh := UpdateNginxResolverService(clientset)
+	<-waitUpdateCh
+	// Delete nginx-resolver deployment and wait until all its pods
```
krisis, line 195: We can relax this since the service is updated to point to the new nginx deployment. Deleting the deployment where k8s eventually deletes the pods that belonging to it allows us to avoid waiting for the pods to be deleted (in the foreground).

krisis (approved):

PR #103, opened 2019-11-12
Fix Add Tenant Command

dvaldivia:


Alevsk (approved):

cluster/context.go
```diff
@@ -150,6 +150,12 @@ func NewContextWithTenantID(tenantID *uuid.UUID) (*Context, error) {
 	return newCtxWithTenant(&tenant), nil
 }
 
+// Creates a new `Context` for a tenant that holds transaction and `context.Context`
+// to control timeouts and cancellations.
+func NewEmpty() (*Context, error) {
```
cesnietor, line 155: I would suggest other name at least to understand that it is a context like `NewEmptyContext`

PR #105, opened 2019-11-13
M3 DB Schema is implicit on connection

dvaldivia:
Before we had to specify the schema and even the db, no more, it's part of the connection

### DIFF
--- a/cluster/migrations/000002_url-tokens.up.sql
+++ b/cluster/migrations/000002_url-tokens.up.sql
@@ -1,0 +1,43 @@
+-- This file is part of MinIO Kubernetes Cloud
+-- Copyright (c) 2019 MinIO, Inc.
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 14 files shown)


cluster/migrations/000002_url-tokens.up.sql
```diff
@@ -0,0 +1,43 @@
...
+    id          uuid
+        constraint url_tokens_pk
+            primary key,
+    tenant_id uuid not null
+        constraint url_tokens_tenant_id_fk
+            references tenants
+            on delete cascade,
+    user_id uuid not null,
+    expiration  timestamptz,
+    used_for    varchar(256),
```
cesnietor, line 28: should we better use a type for `used_for`? so that we let the db validate the use, else we need to validate by strings every time.

cesnietor (approved):

cluster/migrations/000002_url-tokens.up.sql
dvaldivia: Yes, we could introduce an enumerated type on postgres, or all together use a golang type

PR #106, opened 2019-11-13
Add UserInvite grpc handler

cesnietor:
This handler leverages from the current invite user to send an email based on the desired action to make either add new user or reset a user's password.


protos/public_api.proto
```diff
@@ -160,11 +169,32 @@ message LoginResponse {
...
+message SetPasswordRequest {
     string password = 1;
+}
+
+message ChangePasswordRequest {
+    string old_password = 1;
     string new_password = 2;
 }
 
+enum InviteAction {
```
dvaldivia: we should not expose this to APIs, instead it is understood or implicit by the RPC name (SetPassword, ChangePassword)

dvaldivia (changes requested):

protos/public_api.proto
cesnietor: SetPassword is different from Change Password. Set Password is for Signup(Add new user) and Reset Password.

cesnietor: Invitations for both `UserAddInvite` and `UserResetPasswordInvite` in their corresponding handlers

dvaldivia (approved):

PR #107, opened 2019-11-13
Private gRPC API for CLI with Admin Authentication

dvaldivia:


portal/api-private-admin.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	uuid "github.com/satori/go.uuid"
+
+	"github.com/minio/m3/cluster"
+	pb "github.com/minio/m3/portal/stubs"
+)
+
```
cesnietor, line 31: Can you please Add documentation for each handler

portal/api-private-admin.go
```diff
@@ -0,0 +1,60 @@
...
+func (ps *privateServer) AdminAdd(ctx context.Context, in *pb.AdminAddRequest) (*pb.AdminAddResponse, error) {
+	appCtx, err := cluster.NewEmptyContextWithGrpcContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	_, err = cluster.AddAdminAction(appCtx, in.Name, in.Email)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, nil
```
cesnietor: let's return the error similar to public handlers like `nil, status.New(codes.InvalidArgument, "Tenant not valid").Err()`, etc.

portal/api-private-admin.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, status.New(codes.Internal, "Internal error").Err()
+	}
+	tokenID, err := uuid.FromString(in.Token)
+	if err != nil {
+		return nil, status.New(codes.InvalidArgument, "Invalid token").Err()
+	}
+
+	err = cluster.SetAdminPasswordAction(appCtx, &tokenID, in.Password)
+	if err != nil {
+		return &pb.SetAdminPasswordResponse{Error: err.Error()}, nil
```
cesnietor: Same here I think we should deprecate the Error response on the message since else we need to validate the error based on a string instead of an error.

portal/api-private-authentication.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+
+	"github.com/minio/m3/cluster"
+	"golang.org/x/crypto/bcrypt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	pb "github.com/minio/m3/portal/stubs"
+)
+
```
cesnietor, line 29: Same, can you add docs here please?

portal/api-private-authentication.go
```diff
@@ -0,0 +1,69 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cesnietor, line 15: There should be a `\n` after the docs else it will mess up the whole documentation

portal/api-private-tenant.go
```diff
@@ -0,0 +1,34 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cesnietor, line 15: `\n` after the documentation

portal/authentication/authentication.go
```diff
@@ -0,0 +1,61 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cesnietor, line 15: `\n` after the documentation

portal/api-private-tenant.go
```diff
@@ -0,0 +1,34 @@
...
+	pb "github.com/minio/m3/portal/stubs"
+)
+
+func (ps *privateServer) AddTenant(ctx context.Context, in *pb.AddTenantRequest) (*pb.AddTenantResponse, error) {
+	err := cluster.AddTenantAction(in.Name, in.ShortName, in.UserName, in.UserEmail)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, nil
+	}
+	return &pb.AddTenantResponse{Status: "Success"}, nil
```
cesnietor: Same here for status, and errors?

cesnietor (changes requested):

protos/private_api.proto
```diff
@@ -23,6 +23,58 @@ option java_outer_classname = "PrivateApiProto";
...
+    string name = 1;
+    string short_name = 2;
+    string user_name = 3;
+    string user_email = 4;
+}
+
+message AddTenantResponse {
+    string error = 1;
+    string status = 2;
+}
```
cesnietor: should we instead use Empty response?

cluster/admin-sessions.go
```diff
@@ -0,0 +1,123 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cesnietor, line 15: `\n` after doc

cluster/admin-tokens.go
```diff
@@ -0,0 +1,142 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+package cluster
```
cesnietor, line 17: `\n` after doc

portal/api-private-tenant.go
dvaldivia: Success should not return an error, let's discuss about whether to return a message or not

portal/api-private-authentication.go
```diff
@@ -36,12 +38,12 @@ func (ps *privateServer) Login(ctx context.Context, in *pb.CLILoginRequest) (*pb
 	// Look for the user on the database by email
 	admin, err := cluster.GetAdminByEmail(appCtx, in.Email)
 	if err != nil {
-		return nil, status.New(codes.Unauthenticated, "Wrong email and/or password.").Err()
+		return nil, status.New(codes.InvalidArgument, "Wrong email and/or password.").Err()
```
cesnietor: Unauthenticated was ok, since the nature of the error, is there any specific reason to change it?

cmd/m3/portal.go
```diff
@@ -17,18 +17,31 @@
...
 
 // list files and folders.
 var portalCmd = cli.Command{
-	Name:   "portal",
-	Usage:  "starts portal",
-	Action: startAPIPortalCmd,
+	Name:    "service",
+	Aliases: []string{"s"},
+	Usage:   "starts m3 services, public and private APIs.",
+	Action:  startAPIServiceCmd,
```
cesnietor, line 31: Can you please then update the Documentation on the README cause It still relies on portal ?

cesnietor (changes requested):

portal/api-private-authentication.go
dvaldivia: Technically it's not an authentication problem, or the error is not related to that, the error is related to passing incorrect credentials, so it's an argument related problem

portal/api-private-tenant.go
cesnietor: We can probably say that if no error, success?

cmd/m3/portal.go
cesnietor: oh lol nvm it is not there

portal/api-private-tenant.go
```diff
@@ -0,0 +1,36 @@
...
+
+	pb "github.com/minio/m3/portal/stubs"
+)
+
+// AddTenant rpc to add a new tenant and it's first user
+func (ps *privateServer) AddTenant(ctx context.Context, in *pb.AddTenantRequest) (*pb.AddTenantResponse, error) {
+	err := cluster.AddTenantAction(in.Name, in.ShortName, in.UserName, in.UserEmail)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, nil
```
cesnietor: should we return the error?

portal/api-private-authentication.go
```diff
@@ -0,0 +1,71 @@
...
+	"golang.org/x/crypto/bcrypt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	pb "github.com/minio/m3/portal/stubs"
+)
+
+// Login rpc to generate a session for an admin
+func (ps *privateServer) Login(ctx context.Context, in *pb.CLILoginRequest) (*pb.CLILoginResponse, error) {
+	fmt.Println("Login RPC")
```
cesnietor:
```suggestion
```

cesnietor (dismissed):

cesnietor (approved):

PR #116, opened 2019-11-14
CLI Tenant Add. Fix Bug on AddUser.

dvaldivia:
This PR moves the `./m3 tenant add` command to the rpc style of cli calls.

It also fixes a nasty bug that happened because we were doing `QueryRows` on a result set that we expected to contain 1 record inside a transaction. we should keep an eye for this know issue and avoid doing `QueryRows` inside transactions


Alevsk (approved):

cluster/tenant-users.go
```diff
@@ -36,12 +36,8 @@ type User struct {
 // AddUser adds a new user to the tenant's database
 func AddUser(ctx *Context, newUser *User) error {
 	// validate user Name
-	if newUser.Name != "" {
-		// TODO: improve regex
-		var re = regexp.MustCompile(`^[a-zA-Z ]{4,}$`)
-		if !re.MatchString(newUser.Name) {
-			return errors.New("a valid name is needed")
-		}
```
cesnietor, line 44: we are not validating the name ? where are we doing it?

cluster/tenants.go
```diff
@@ -121,9 +130,9 @@ func AddTenantAction(name, shortName, adminName, adminEmail string) error {
 		return err
 	}
 	// if the first admin name and email was provided send them an invitation
```
cesnietor, line 132:
```suggestion
	// if the first user name and email was provided send them an invitation
```

cluster/tenants.go
```diff
@@ -50,15 +50,24 @@ const (
 	BucketCustom
 )
 
-// AddTenantAction adds a tenant to the cluster, if an admin name and email are provided, the user is created and invited
+// TenantAddAction adds a tenant to the cluster, if an admin name and email are provided, the user is created and invited
```
cesnietor, line 53: I like to start doing this naming convention.

cmd/m3/tenant-add.go
```diff
@@ -88,23 +88,36 @@ func addTenant(ctx *cli.Context) error {
...
 	}
 
-	if name != "" && email == "" {
-		fmt.Println("User email is needed")
+	if userName == "" || userEmail == "" {
+		fmt.Println("User name and email is needed")
 		return errMissingArguments
 	}
 
-	err := cluster.AddTenantAction(tenantName, tenantShortName, name, email)
```
cesnietor, line 105: Since this will change the way we create Tenants we will need to update documentation and add for Operator otherwise it will be broken for development. Or should we still allow it from m3?

PR #118, opened 2019-11-14
CLI Admin Add via RPC

dvaldivia:


cmd/m3/admin-add.go
```diff
@@ -67,20 +67,24 @@ func adminAdd(ctx *cli.Context) error {
...
+		Name:  name,
+		Email: email,
+	})
 
-	admin, err := cluster.AddAdminAction(apptCtx, name, email)
 	if err != nil {
-		fmt.Println("Error adding user:", err.Error())
-		return err
+		fmt.Println(err)
+		return nil
```
cesnietor, line 84: error handling?

dvaldivia: This will print the error for the users to understand why the request failed

cesnietor (approved):

PR #119, opened 2019-11-14
CLI Set Admin Password via RPC

dvaldivia:


cmd/m3/set-password.go
```diff
@@ -0,0 +1,108 @@
...
+	defer cnxs.Conn.Close()
+	// perform RPC
+	_, err = cnxs.Client.SetPassword(cnxs.Context, &pb.SetAdminPasswordRequest{
+		Password: password,
+		Token:    token,
+	})
+
+	if err != nil {
+		fmt.Println(err)
+		return nil
```
cesnietor, line 102: error handling?

cesnietor (approved):

cmd/m3/set-password.go
dvaldivia: this will print the error on the terminal so the user can understand what happened

PR #120, opened 2019-11-14
CLI Cluster Storage Group Add via RPC

dvaldivia:


portal/api-private-cluster.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"fmt"
+
+	"github.com/minio/m3/cluster"
+	pb "github.com/minio/m3/portal/stubs"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// AddTenant rpc to add a new tenant and it's first user
```
cesnietor: doc update?

portal/api-private-cluster.go
```diff
@@ -0,0 +1,57 @@
...
+func (ps *privateServer) ClusterScSgAdd(ctx context.Context, in *pb.StorageGroupAddRequest) (*pb.StorageGroupAddResponse, error) {
+	appCtx, err := cluster.NewEmptyContext()
+	if err != nil {
+		return nil, status.New(codes.Internal, "An internal error happened").Err()
+	}
+
+	var name *string
+	if in.Name != "" {
+		name = &in.Name
+	}
```
cesnietor, line 39: `name := in.GetName()` it does pretty much what you are doing.

cesnietor (changes requested):

portal/api-private-cluster.go
dvaldivia: I need an empty reference if the name was not passed `in.Name` and `in.GetName()` will return "" if the string was not passed

portal/api-private-cluster.go
dvaldivia: done

cesnietor (approved):

cesnietor (approved):

PR #128, opened 2019-11-15
Update list users grpc handler

cesnietor:


protos/public_api.proto
```diff
@@ -299,6 +299,7 @@ message User {
     string name = 1;
     string email = 2;
     string id = 3;
+    bool is_me = 4;
```
dvaldivia, line 302: can we call it `its_a_me` ?

dvaldivia (approved):

PR #136, opened 2019-11-16
CLI Dev Command

dvaldivia:
Introduces a Development command called `./m3 dev` which will do the following

1. Start monitoring m3/nginx pods
2. Port forward nginx (80), public api (50051) and private api (50052) 
3. if a pod gets created (nginx redeployed, m3 redeployed) the port-forwards are restarted

We could potentially also add postgres port-forwarding by default

### DIFF
--- a/cmd/m3/dev.go
+++ b/cmd/m3/dev.go
@@ -1,0 +1,273 @@
+// This file is part of MinIO Kubernetes Cloud
+// Copyright (c) 2019 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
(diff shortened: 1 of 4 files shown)


cmd/m3/dev.go
```diff
@@ -0,0 +1,260 @@
...
+	m3PFCtx, m3Cancel := context.WithCancel(context.Background())
+	nginxCtx, nCancel := context.WithCancel(context.Background())
+	doneCh := make(chan struct{})
+
+	//go func() {
+	//	time.Sleep(time.Second * 10)
+	//	fmt.Println("Simulate cancellation")
+	//	close(doneCh)
+	//	m3Cancel()
+	//}()
```
cesnietor, line 70: Can you please remove unused comments

cmd/m3/dev.go
```diff
@@ -0,0 +1,260 @@
...
+					nCancel()
+					// restart the context
+					nginxCtx, nCancel = context.WithCancel(context.Background())
+				}
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			pod := obj.(*v1.Pod)
+			if strings.HasPrefix(pod.ObjectMeta.Name, "m3") {
+				fmt.Println("An m3 pod was deleted:", pod.ObjectMeta.Name)
```
cesnietor, line 141: we don't close the channel after a deletion.

dvaldivia: no, we wait instead for the creation, this delete print is for informative purposes :P

cesnietor (approved):

PR #138, opened 2019-11-18
Rename Portal Package to API

dvaldivia:
We are using the `portal` package more like an `api` package, so I'm renaming it so we can use portal for the actual portal backend


protos/private_api.proto
```diff
@@ -47,6 +47,9 @@ service PrivateAPI {
     }
     rpc Login (CLILoginRequest) returns (CLILoginResponse) {
     }
+    // Session related
+    rpc ValidateSession (AdminEmpty) returns (AdminEmpty) {
```
cesnietor, line 51: Sneaky

cesnietor (approved):

PR #149, opened 2019-11-21
Tenant sa perms

dvaldivia:
Based on @nitisht 's branch, this PR introduces `./tenant permission add` which create a permission with multiple resources and actions

for example, to grant `write` permission to `bucketA` and `bucketB` 

`./m3 tenant permission add acme SAMPLE allow write "bucketA,bucketB"`


nitisht (approved):
LGTM

protos/private_api.proto
```diff
@@ -76,6 +79,17 @@ message TenantBucketAddRequest {
 message TenantBucketAddResponse {
 }
 
+message TenantPermissionAddRequest {
```
cesnietor, line 82: Shouldn't this be in the public proto? or both?

cesnietor (approved):

PR #154, opened 2019-11-24
Protect Public API With gRPC Interceptor

dvaldivia:
Protects all the inbound grpc calls with an interceptor.

Also removes deprecated context functions


dvaldivia:
I tested this change with the UI making calls into the grpc and everything seems to work fine, I only created an exception for 3 calls

api/authentication/authentication.go
```diff
@@ -64,3 +67,51 @@ func AdminAuthInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
...
+func PublicAuthInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	// exempted calls from the validation
+	if info.FullMethod == "/m3.PublicAPI/Login" ||
+		info.FullMethod == "/m3.PublicAPI/SetPassword" ||
+		info.FullMethod == "/m3.PublicAPI/ValidateInvite" {
+		// log this call
+		log.Printf("%s", info.FullMethod)
+		return handler(ctx, req)
+	}
+	fmt.Println(info.FullMethod)
```
cesnietor:
```suggestion
```

api/authentication/authentication.go
```diff
@@ -64,3 +67,51 @@ func AdminAuthInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
...
+		log.Println(err)
+		return nil, status.New(codes.Internal, "internal error").Err()
+	}
+
+	// attach the details of the session to the context
+	ctx = context.WithValue(ctx, cluster.SessionIDKey, validSession.ID)
+	ctx = context.WithValue(ctx, cluster.UserIDKey, validSession.UserID.String())
+	ctx = context.WithValue(ctx, cluster.TenantIDKey, validSession.TenantID.String())
+	ctx = context.WithValue(ctx, cluster.TenantShortNameKey, tenant.ShortName)
+	ctx = context.WithValue(ctx, cluster.WhoAmIKey, user.Email)
```
Alevsk, line 110: is the user email is being used as a user identifier at some point during the authentication flow? if yes can we change this to something more obscure such as auto generated id?

dvaldivia: it's not, this user email is being pulled from the session, every call we attach the email to the context so we can track who is doing what or who is doing the action, in short: token -> session -> user

Alevsk (approved):

PR #159, opened 2019-11-25
Permission-api-public-handler, Add Update handler to update a permiss…

cesnietor:
…ion with it's resources and actions
Update api request should be as a PUT `/api/v1/permissions/{id}`
And send the values as a normal Create, Backend will do the update accordingly.


api/api-private-tenant-permissions.go
```diff
@@ -56,7 +56,7 @@ func (s *privateServer) TenantPermissionAdd(ctx context.Context, in *pb.TenantPe
 	}
 	appCtx.Tenant = &tenant
 
-	if _, err := cluster.AddPermission(appCtx, in.Name, in.Description, effect, in.Resources, in.Actions); err != nil {
+	if _, err := cluster.AddPermissionToDB(appCtx, in.Name, in.Description, effect, in.Resources, in.Actions); err != nil {
```
dvaldivia, line 59: technically this adds the permission to the DB and also commits the change to MinIO

api/api-public-tenant-permissions.go
```diff
@@ -48,34 +48,8 @@ func (s *server) ListPermissions(ctx context.Context, in *pb.ListPermissionsRequ
...
-		for _, permAction := range perm.Actions {
-			pbAction := pb.PermissionAction{
-				Id:   permAction.ID.String(),
-				Type: string(permAction.ActionType),
-			}
-			pbPerm.Actions = append(pbPerm.Actions, &pbAction)
-		}
-
-		pbPerms = append(pbPerms, &pbPerm)
+		pbPerm := buildPermissionResponseFromPermissionObj(perm)
```
dvaldivia: maybe something like`buildPermissionPBFromPermission`

api/api-public-tenant-permissions.go
```diff
@@ -130,37 +103,164 @@ func (s *server) AddPermission(ctx context.Context, in *pb.AddPermissionRequest)
...
+	}
+
+	// get updated permission
+	updatedPermission, err := cluster.GetPermissionByID(appCtx, permission.ID.String())
+	if err != nil {
+		return nil, status.New(codes.InvalidArgument, "permission not found").Err()
+	}
+	// Create response object
+	permissionResponse := buildPermissionResponseFromPermissionObj(updatedPermission)
+
```
dvaldivia: lastly, you need to see which service accounts use this permission and update their policies:
```
// update the policy for each SA
	var saChs []chan error
	for _, sa := range serviceAccountIDs {
		ch := UpdatePolicyForServiceAccount(ctx, sgt.StorageGroupTenant, tenantConf, sa)
		saChs = append(saChs, ch)
	}
	// wait for all to finish
	for _, ch := range saChs {
		err := <-ch
		if err != nil {
			return err
		}
	}
```

dvaldivia (changes requested):

api/api-private-tenant-permissions.go
cesnietor: ooki

api/api-public-tenant-permissions.go
```diff
@@ -130,37 +103,164 @@ func (s *server) AddPermission(ctx context.Context, in *pb.AddPermissionRequest)
...
+	// TODO: parallelize
+	resourcesToCreate := differenceArrays(resourcesBucketNames, currentResourceBucketNames)
+	resourcesToDelete := differenceArrays(currentResourceBucketNames, resourcesBucketNames)
+	// CREATE New Resources
+	// create a Temporal permission to Create the new Permission resources
+	tempPermission := &cluster.Permission{ID: permission.ID}
+	cluster.AppendPermissionResourcesObj(tempPermission, resourcesToCreate)
+	// for each resource, save to DB
+	for _, resc := range tempPermission.Resources {
+		err = cluster.InsertResource(appCtx, tempPermission, &resc)
```
Alevsk: Do we support bulk insert instead of creating the records one by one?

api/api-public-tenant-permissions.go
```diff
@@ -130,37 +103,164 @@ func (s *server) AddPermission(ctx context.Context, in *pb.AddPermissionRequest)
...
+			return nil, err
+		}
+	}
+	// DELETE unwanted Resources
+	for _, bucketName := range resourcesToDelete {
+		resourceID, ok := mapResourceName[bucketName]
+		if !ok {
+			return nil, status.New(codes.Internal, "error retrieving resource to delete").Err()
+		}
+		err = cluster.DeletePermissionResourceDB(appCtx, resourceID)
```
Alevsk: Do we support bulk delete instead of deleting the records one by one?

api/api-public-tenant-permissions.go
```diff
@@ -130,37 +103,164 @@ func (s *server) AddPermission(ctx context.Context, in *pb.AddPermissionRequest)
...
 	}
-	return res
+	// TODO: parallelize
+	actionsToCreate := differenceArrays(actionTypes, currentActionTypes)
+	actionsToDelete := differenceArrays(currentActionTypes, actionTypes)
+	// CREATE New Actions
+	cluster.AppendPermissionActionObj(tempPermission, actionsToCreate)
+	// for each resource, save to DB
+	for _, action := range tempPermission.Actions {
+		err = cluster.InsertAction(appCtx, tempPermission, &action)
```
Alevsk: Can we do bulk insert for permission actions instead of inserting them one by one?

api/api-public-tenant-permissions.go
```diff
@@ -130,37 +103,164 @@ func (s *server) AddPermission(ctx context.Context, in *pb.AddPermissionRequest)
...
+			return nil, err
+		}
+	}
+	// DELETE unwanted Actions
+	for _, types := range actionsToDelete {
+		actionID, ok := mapActionTypes[types]
+		if !ok {
+			return nil, status.New(codes.Internal, "error retrieving action to delete").Err()
+		}
+		err = cluster.DeletePermissionActionDB(appCtx, actionID)
```
Alevsk: Do we support bulk delete instead of deleting the records one by one?

api/api-public-tenant-permissions.go
dvaldivia: maybe for resources it makes sense

api/api-public-tenant-permissions.go
cesnietor: done please review

api/api-public-tenant-permissions.go
cesnietor: there is a way but so far it requires to do a COPY in postgres which requires a file, I think we can come up with an improvement later on.

api/api-public-tenant-permissions.go
cesnietor: Done

api/api-public-tenant-permissions.go
cesnietor: we'll check in further iterations since so far there are not that many inserts for MVP

api/api-public-tenant-permissions.go
cesnietor: done

api/api-public-tenant-permissions.go
cesnietor: done

api/api-public-tenant-permissions.go
Alevsk: the other approach is to create a query string like `INSERT INTO films (code, title, did, date_prod, kind) VALUES` and then for each object to be inserted generate and append to the query a string like this `(?, ?, ?, ?, ?),`, finally flat all the objects attributes into a single array and pass that array to the query exec, its efficient but we need to be very careful since in most languages you cannot trust the object attributes ordering.

Alevsk (approved):

dvaldivia (approved):

PR #161, opened 2019-11-26
Rename path field to pattern on permission_resources

dvaldivia:


cluster/tenant-migrations/000001_tenant-initial.up.sql
```diff
@@ -104,7 +104,7 @@ CREATE TABLE permissions_resources
             REFERENCES permissions
             ON DELETE CASCADE,
     bucket_name      VARCHAR(64)                            NOT NULL,
-    path             VARCHAR(512)                           NOT NULL,
+    pattern          VARCHAR(512)                           NOT NULL,
```
cesnietor, line 107: for the next prs we need new migrations instead of changing the first one.

cesnietor (approved):

cluster/tenant-migrations/000001_tenant-initial.up.sql
dvaldivia: yeah, we need the `migrate` or `upgrade` command for that

PR #168, opened 2019-11-27
Store Email Templates on the Database

dvaldivia:
introduces cli command `./m3 email-template set <TEMPLATE_ID> <TEMPLATE_BODY>` to set email to a desired custom implementation


protos/private_api.proto
```diff
@@ -69,6 +69,9 @@ service PrivateAPI {
     // Cluster Operations
     rpc ClusterScSgAdd (StorageGroupAddRequest) returns (StorageGroupAddResponse) {
     }
+    // Configuration operations
```
cesnietor:
```suggestion
    // Sets Email Template
```

cluster/email.go
```diff
@@ -126,20 +128,90 @@ func SendMail(toName, toEmail, subject, body string) error {
...
+	err := row.Scan(&template)
+	if err != nil {
+		return nil, err
+	}
+	return &template, nil
+}
+
+// SetEmailTemplate upserts a template into the database. If the id is not present the record will be inserted, if it's
+// present it will be updated
+func SetEmailTemplate(ctx *Context, templateName, templateBody string) error {
```
cesnietor, line 206: templateID is being used not the name
```suggestion
func SetEmailTemplate(ctx *Context, templateID, templateBody string) error {
```

cluster/email.go
```diff
@@ -126,20 +128,90 @@ func SendMail(toName, toEmail, subject, body string) error {
...
+					email_templates (id, template) 
+				VALUES ($1, $2) 
+				ON CONFLICT (id) DO 
+			    UPDATE SET template=$2`
+	tx, err := ctx.MainTx()
+	if err != nil {
+		return err
+	}
+	// Execute query
+	_, err = tx.Exec(query, templateName, templateBody)
```
cesnietor, line 223: template ID?

cluster/migrations/000004_email-templates.up.sql
```diff
@@ -0,0 +1,29 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+CREATE TABLE email_templates
+(
+    id       VARCHAR(256) NOT NULL
```
cesnietor: we should use either `id` type UUID or `id` type UUID and `name` as other column to be consistent

cluster/email.go
```diff
@@ -126,20 +128,90 @@ func SendMail(toName, toEmail, subject, body string) error {
...
+			return nil, err
+		}
+	} else {
+		// if we found no template, load from disk
+		// Working Directory
+		wd, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+		t, err = template.ParseFiles(wd + fmt.Sprintf("/cluster/templates/%s.html", templateName))
```
Alevsk, line 160: does the user controls the templateName value in any way? what if he does `../../home/root/.ssh/id_rsa;` 

alt text

We need to add some validations:

- Reject file names which contain ., .. or / (or \ under Windows)
- Limit file names to basic alphanumeric characters
- Prepend the include directory name and append the appropriate extension

cluster/migrations/000004_email-templates.up.sql
dvaldivia: the thing is that the template will have a unique key based on name, if we rely on `id` it will be an unused field since we will only reference the template by their name, therefore the name should be it's primary key

cluster/email.go
dvaldivia: there's other files inside the container, other than the ca certs and other templates

tenor (1)

cluster/migrations/000004_email-templates.up.sql
cesnietor: we should then reference it by id and used the name just when querying it  similiar to what we do on ServiceAccounts and Permissions

Alevsk (dismissed):

protos/private_api.proto
```diff
@@ -189,3 +192,10 @@ message TenantServiceAccountAssignRequest {
 }
 message TenantServiceAccountAssignResponse {
 }
+
+message SetEmailTemplateRequest {
+    string id = 1;
```
cesnietor: should this be also name?

cluster/email.go
```diff
@@ -126,20 +129,100 @@ func SendMail(toName, toEmail, subject, body string) error {
...
+	err := row.Scan(&template)
+	if err != nil {
+		return nil, err
+	}
+	return &template, nil
+}
+
+// SetEmailTemplate upserts a template into the database. If the id is not present the record will be inserted, if it's
+// present it will be updated
+func SetEmailTemplate(ctx *Context, templateID, templateBody string) error {
```
cesnietor:
```suggestion
func SetEmailTemplate(ctx *Context, templateName, templateBody string) error {
```
?

cmd/m3/email-template-set.go
```diff
@@ -0,0 +1,86 @@
...
+		cli.StringFlag{
+			Name:  "template",
+			Value: "",
+			Usage: "template body",
+		},
+	},
+}
+
+func emailTemplateSet(ctx *cli.Context) error {
+	id := ctx.String("id")
```
Alevsk: should we change all this ids to name?

dvaldivia: done

protos/private_api.proto
dvaldivia: done

cesnietor (approved):

PR #170, opened 2019-11-27
Service account enable disable grpc public handler

cesnietor:
Enable/disable support for Service Accounts, also modifies Minio User Status
`/api/v1/service_accounts/{id}/enable`
`/api/v1/service_accounts/{id}/disable`

### DIFF
--- a/api/api-public-tenant-service-account.go
+++ b/api/api-public-tenant-service-account.go
@@ -246,7 +246,7 @@ func (s *server) EnableServiceAccount(ctx context.Context, in *pb.ServiceAccount
 		err = cluster.SetMinioServiceAccountStatus(appCtx, serviceAccount, true)
 		if err != nil {
 			log.Println(err.Error())
-			return nil, status.New(codes.NotFound, "Service Account Not Found").Err()
+			return nil, status.New(codes.Internal, "Error Updating Status").Err()
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


api/api-public-tenant-service-account.go
```diff
@@ -195,6 +209,103 @@ func (s *server) UpdateServiceAccount(ctx context.Context, in *pb.UpdateServiceA
...
+	defer func() {
+		if err != nil {
+			appCtx.Rollback()
+			return
+		}
+		err = appCtx.Commit()
+	}()
+
+	// Fetch the service account
+	serviceAccount, err := cluster.GetServiceAccountByID(appCtx, &id)
```
dvaldivia, line 238: if you put this on a function I can reuse it from CLI

api/api-public-tenant-service-account.go
```diff
@@ -195,6 +209,103 @@ func (s *server) UpdateServiceAccount(ctx context.Context, in *pb.UpdateServiceA
...
```
dvaldivia, line 287: Same as above

cesnietor: is already really abstracted, you can use it in the same way from cli and then using `SetMinioServiceAccountStatus`

dvaldivia (approved):

PR #174, opened 2019-12-02
Remove hardcoded Hosts and Disk volumes

dvaldivia:
Removes the hardcoded 4 hosts and 4 disks. 

Introduces all the commands to support this, like adding nodes and storage clusters, have them being associated and such.

### DIFF
--- a/api/api-private-cluster-nodes.go
+++ b/api/api-private-cluster-nodes.go
@@ -1,0 +1,227 @@
+// This file is part of MinIO Kubernetes Cloud
+// Copyright (c) 2019 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package api
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"regexp"
+	"strconv"
+	"strings"
+
+	uuid "github.com/satori/go.uuid"
+
+	"github.com/minio/m3/cluster"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	pb "github.com/minio/m3/api/stubs"
+)
+
+// ClusterNodesAdd rpc to add a new node to the cluster
+func (ps *privateServer) ClusterNodesAdd(ctx context.Context, in *pb.NodeAddRequest) (*pb.NodeAddResponse, error) {
+	if in.Name == "" {
+		return nil, status.New(codes.InvalidArgument, "A node name is needed").Err()
+	}
+
+	if in.K8SLabel == "" {
+		return nil, status.New(codes.InvalidArgument, "A kubernetes label is needed").Err()
+	}
+
+	// validate volumes string
+	if in.Volumes == "" {
+		return nil, status.New(codes.InvalidArgument, "A list of volumes on the node is needed").Err()
+	}
+
+	var volumes []string
+
+	if strings.Contains(in.Volumes, "...") {
+		var re = regexp.MustCompile(`^.*?({([0-9]+)\.\.\.([0-9]+)}).*?$`)
+		if !re.MatchString(in.Volumes) {
+			return nil, status.New(codes.InvalidArgument, "Invalid descriptor of volumes mount points").Err()
+		} else {
+			rs := re.FindStringSubmatch(in.Volumes)
+			// parse nums, regex should have validated valid integers
+			fromNum, _ := strconv.Atoi(rs[2])
+			toNum, _ := strconv.Atoi(rs[3])
+			// if we received /mnt/disk{1...4} we will iterate 4 times, and replace {1...4} for the corresponding number
+			for i := fromNum; i <= toNum; i++ {
+				volumes = append(volumes, strings.Replace(in.Volumes, rs[1], fmt.Sprintf("%d", i), 1))
+			}
+		}
+
+	} else {
+		// attempt to validate single volume mount path
+		randomNodeID := uuid.NewV4()
+		if _, err := cluster.NewVolume(&randomNodeID, in.Volumes); err != nil {
+			return nil, status.New(codes.InvalidArgument, "Volume mount path is not valid").Err()
+		}
+		volumes = append(volumes, in.Volumes)
+	}
+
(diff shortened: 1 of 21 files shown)


api/api-private-cluster-nodes.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// validate volumes string
+	if in.Volumes == "" {
+		return nil, status.New(codes.InvalidArgument, "A list of volumes on the node is needed").Err()
+	}
+
+	var volumes []string
+
+	if strings.Contains(in.Volumes, "...") {
+		var re = regexp.MustCompile(`^.*?({([0-9]+)\.\.\.([0-9]+)}).*?$`)
```
harshavardhana: You should use pkg/ellipses package from minio/minio

harshavardhana (changes requested):

cluster/nodes.go
```diff
@@ -0,0 +1,228 @@
...
+		WHERE 
+		      sg.id = $1`
+
+	rows, err := GetInstance().Db.Query(queryNodes, storageGroupID)
+	if err != nil {
+		return nil, err
+	}
+	var nodes []*StorageGroupNode
+	nodeMap := make(map[uuid.UUID]*Node)
+	var nodeIDs []uuid.UUID
```
dvaldivia, line 192: rows.Close() -Cesar

cluster/nodes.go
```diff
@@ -0,0 +1,228 @@
...
+	node := Node{}
+	err := row.Scan(&node.ID, &node.Name, &node.K8sLabel)
+	if err != nil {
+		return nil, err
+	}
+
+	return &node, nil
+}
+
+type Volume struct {
```
dvaldivia: can we rename this to NodeVolume so that it is consistent with the node_volumes table -Cesar

cluster/nodes.go
```diff
@@ -0,0 +1,228 @@
...
+		FROM 
+			node_volumes nv
+		WHERE 
+		      nv.node_id = ANY($1)`
+
+	volRows, err := GetInstance().Db.Query(queryVolumes, pq.Array(nodeIDs))
+	if err != nil {
+		return nil, err
+	}
+	for volRows.Next() {
```
dvaldivia, line 223: rows.Close() -Cesar

api/api-private-cluster-nodes.go
```diff
@@ -0,0 +1,225 @@
...
+		// parse nums, regex should have validated valid integers
+		fromNum, _ := strconv.Atoi(rs[2])
+		toNum, _ := strconv.Atoi(rs[3])
+		// if we received /mnt/disk{1...4} we will iterate 4 times, and replace {1...4} for the corresponding number
+		for i := fromNum; i <= toNum; i++ {
+			volumes = append(volumes, strings.Replace(in.Volumes, rs[1], fmt.Sprintf("%d", i), 1))
+		}
+	} else {
+		// attempt to validate single volume mount path
+		randomNodeID := uuid.NewV4()
```
cesnietor: can we put this inside cluster.NewVolume() similar to the way you already do it on `NewNode()`

api/api-private-cluster-nodes.go
dvaldivia: ready

api/api-private-cluster-nodes.go
```diff
@@ -0,0 +1,225 @@
...
+		if _, err := cluster.NewVolume(&randomNodeID, in.Volumes); err != nil {
+			return nil, status.New(codes.InvalidArgument, "Volume mount path is not valid").Err()
```
cesnietor: we are using a specific error related to all errors that can happen during `NewVolume()` we should probably use a more generic message

api/api-private-cluster-nodes.go
dvaldivia: This is merely to use `NewVolume` for validation, it needs some `nodeID` so I'm randomly generating one.

api/api-private-cluster-nodes.go
dvaldivia: This is merely to use NewVolume for validation, it needs some nodeID so I'm randomly generating one.

dvaldivia: A more clear error message was added

Alevsk (dismissed):

cesnietor:
following the README:
```
➜ ./m3 cluster sc sg add --name my-dc-rack-1
rpc error: code = InvalidArgument desc = Invalid storage cluster name.
```

dvaldivia:
@cesnietor that's an outdated command, the new one is in the reame

cluster/nodes.go
```diff
@@ -0,0 +1,230 @@
...
+		SELECT 
+				n.id,n.name,n.k8s_label,scn.num
+		FROM 
+			nodes n
+		LEFT JOIN storage_cluster_nodes scn ON n.id = scn.node_id
+		LEFT JOIN storage_groups sg ON scn.storage_cluster_id = sg.storage_cluster_id
+		WHERE 
+		      sg.id = $1`
+
+	rows, err := GetInstance().Db.Query(queryNodes, storageGroupID)
```
cesnietor: This needs to be like this else it won't get the storage groups created in the earlier steps since they haven't been committed.
```suggestion
	rows, err := ctx.mainTx.Query(queryNodes, storageGroupID)
```

cluster/nodes.go
```diff
@@ -0,0 +1,230 @@
...
+
+	queryVolumes := `
+		SELECT 
+				nv.node_id, nv.id, nv.mount_path, nv.num
+		FROM 
+			node_volumes nv
+		WHERE 
+		      nv.node_id = ANY($1)`
+
+	volRows, err := GetInstance().Db.Query(queryVolumes, pq.Array(nodeIDs))
```
cesnietor: This needs to be like this else it won't get the storage groups created in the earlier steps since they haven't been committed.
```suggestion
	volRows, err := ctx.mainTx.Query(queryVolumes, pq.Array(nodeIDs))
```

cmd/m3/cluster-sc-add.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	defer cnxs.Conn.Close()
+	// perform RPC
+	if _, err = cnxs.Client.ClusterStorageClusterAdd(cnxs.Context, &pb.StorageClusterAddRequest{
+		Name: name,
+	}); err != nil {
+		fmt.Println(err)
+		return nil
+	}
+	fmt.Println("Done adding storage group")
```
cesnietor:
```suggestion
	fmt.Println("Done adding storage cluster")
```

cesnietor (changes requested):

cluster/nodes.go
dvaldivia: we should never reference the private mainTx directly, instead use MainTx() which will ensure a transaction is started if none has been

dvaldivia:
@harshavardhana Ellipses has been implemented

dvaldivia:
@cesnietor Implemented the MainT()

cesnietor (approved):
LGTM

harshavardhana (approved):

PR #177, opened 2019-12-04
Wait for MinIO to be ready on TenantAdd

dvaldivia:
Introduces a function that pings a Tenant's MinIO to make sure it's ready before it proceeds

### DIFF
--- a/cluster/minio.go
+++ b/cluster/minio.go
@@ -112,3 +112,19 @@ func RemoveMinioUser(sgt *StorageGroupTenant, tenantConf *TenantConfiguration, u
 func tagErrorAsMinio(err error) error {
 	return fmt.Errorf("MinIO: %s", err.Error())
 }
+
+// minioIsReady determines whether the MinIO for a tenant is ready or not
+func minioIsReady(ctx *Context) (bool, error) {
+	// Get tenant specific MinIO client
+	minioClient, err := newTenantMinioClient(ctx, ctx.Tenant.ShortName)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = minioClient.ListBuckets()
+	if err != nil {
+		return false, tagErrorAsMinio(err)
+	}
+
+	return true, nil
+}
--- a/cluster/minio.go
+++ b/cluster/minio.go
@@ -120,8 +120,10 @@ func minioIsReady(ctx *Context) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-
-	_, err = minioClient.ListBuckets()
+	// Generate a random bucket name
+	randBucket := RandomCharString(32)
+	// Check if it exist, we expect it to say no, or fail if MinIO is not ready
+	_, err = minioClient.BucketExists(randBucket)
 	if err != nil {
 		return false, tagErrorAsMinio(err)
 	}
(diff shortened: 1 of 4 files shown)


cluster/minio.go
```diff
@@ -112,3 +112,19 @@ func RemoveMinioUser(sgt *StorageGroupTenant, tenantConf *TenantConfiguration, u
 func tagErrorAsMinio(err error) error {
 	return fmt.Errorf("MinIO: %s", err.Error())
 }
+
+// minioIsReady determines whether the MinIO for a tenant is ready or not
+func minioIsReady(ctx *Context) (bool, error) {
+	// Get tenant specific MinIO client
+	minioClient, err := newTenantMinioClient(ctx, ctx.Tenant.ShortName)
```
harshavardhana, line 119: You should just use /minio/healthcheck/live 

```
func testLivenessEndpoint(endpoint string) {
	startTime := time.Now()
	function := "testLivenessEndpoint"

	u, err := url.Parse(fmt.Sprintf("%s%s", endpoint, livenessPath))
	if err != nil {
		// Could not parse URL successfully
		failureLog(function, nil, startTime, "", "URL Parsing for Healthcheck Liveness handler failed", err).Fatal()
	}

	tr := &http.Transport{
		TLSClientConfig: &tls.Config{InsecureSkipVerify: u.Scheme == "https"},
	}
	client := &http.Client{Transport: tr, Timeout: timeout}
	resp, err := client.Get(u.String())
	if err != nil {
		// GET request errored
		failureLog(function, nil, startTime, "", "GET request failed", err).Fatal()
	}
	if resp.StatusCode != http.StatusOK {
		// Status not 200 OK
		failureLog(function, nil, startTime, "", fmt.Sprintf("GET /minio/health/live returned %s", resp.Status), err).Fatal()
	}

	defer resp.Body.Close()
	defer successLogger(function, nil, startTime).Info()
}
```

It doesn't need an authenticated call - also doing ListBuckets() might get expensive.

harshavardhana (changes requested):

cluster/tenants.go
```diff
@@ -120,6 +121,20 @@ func TenantAddAction(ctx *Context, name, shortName, userName, userEmail string)
 	}
 	// if the first admin name and email was provided send them an invitation
 	if userName != "" && userEmail != "" {
+		// wait for MinIO to be ready before creating the first user
+		for {
```
cesnietor, line 127: can we allow a number of trial times or so?
we should not allow the possibility of infinite loops since it could possibly never be ready.

cluster/minio.go
dvaldivia: I think liveness is not going to be enough, we need MinIO to be ready to accept new users or configuration changes, if liveness is reported but we cannot make `mc admin user add` then it's not good, we need MinIO to be ready to take regular requests, not sure if this is what readiness reports on

If we have a 4 node cluster, how many nodes do we need online before we can add a user or change a configuration?

dvaldivia: Ready, using Head Bucket

cluster/tenants.go
dvaldivia: implmented

harshavardhana (approved):

cesnietor (approved):

PR #178, opened 2019-12-05
nginx-configmap - adding portal route to nginx resolver

Alevsk:


cluster/nginx.go
```diff
@@ -250,10 +250,25 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
 																		'$status $body_bytes_sent "$http_referer" '
 																		'"$http_user_agent" "$http_x_forwarded_for"';
 				server {
-					#listen 80 default_server;
-					#listen 443 ssl default_server;
-					server_name _ ;
-					return 404;
+					location / {
```
dvaldivia, line 253: what about /api/ ?

dvaldivia (changes requested):

cluster/nginx.go
```diff
@@ -250,10 +250,25 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
...
+						 proxy_set_header X-Frame-Options SAMEORIGIN;
+						 proxy_buffers 256 16k;
+						 proxy_buffer_size 16k;
+						 client_body_timeout 60;
+						 send_timeout 300;
+						 lingering_timeout 5;
+						 proxy_connect_timeout 90;
+						 proxy_send_timeout 300;
+						 proxy_read_timeout 90s;
+						 proxy_pass http://faceliftfe:80;
```
dvaldivia: can we call the service `frontend` ?

dvaldivia (changes requested):

cluster/nginx.go
dvaldivia: `portal`

dvaldivia: `app-router`

k8s/deployments/m3-portal-backend-deployment.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
dvaldivia, line 1: I think you can do this from `m3 setup`

k8s/deployments/m3-portal-proxy.yaml
```diff
@@ -0,0 +1,107 @@
+---
```
dvaldivia, line 1: I think you should do this from `m3 setup`

dvaldivia (changes requested):

cluster/nginx.go
Alevsk: This is the nginx.conf file of `nginx-resolver`, we don't need to handle `/api/` here, request are going to be forwarded to `http://m3-portal-proxy:80`, that proxy load balancer knows to route traffic between frontend and backend.

dvaldivia (approved):

PR #181, opened 2019-12-07
Bucket notification postgres

cesnietor:
Set postgres notification config on tenant creation and during bucket creation.
-Creates a table `bucketevents` on m3 database  and registers all bucket events.

### DIFF
--- a/cluster/credentials.go
+++ b/cluster/credentials.go
@@ -19,6 +19,8 @@ package cluster
 import (
 	"errors"
 	"fmt"
+	"log"
+	"os"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


cluster/credentials.go
```diff
@@ -76,6 +75,11 @@ func createUserCredentials(ctx *Context, tenantShortName string, userdID uuid.UU
 	if err != nil {
 		return err
 	}
+	// create minio postgres configuration for bucket notification
```
dvaldivia: Shouldn't we create this on a per tenant basis instead?

dvaldivia (changes requested):

cluster/credentials.go
cesnietor: true, done.

dvaldivia (approved):

PR #189, opened 2019-12-10
[WIP] Delete tenant , deprovisioning of a tenant by removing all tenant's related data.

cesnietor:
When deleting a tenant it will will create a job for removing the tenant's mounts `/mnt/tdisk/<tenant>` and will redeploy all hosts.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -305,6 +305,26 @@ func DeprovisionTenantOnStorageGroup(ctx *Context, tenant *Tenant, sg *StorageGr
 			ch <- sgTenantResult.Error
 			return
 		}
+		ctx.Tenant = sgTenantResult.StorageGroupTenant.Tenant
+		// Get the credentials for a tenant
+		tenantConf, err := GetTenantConfig(tenant)
+		if err != nil {
+			ch <- err
+			return
+		}
+		// Unset Postgres notification before deleting tenant
+		err = unsetMinioConfigPostgresNotification(sgTenantResult.StorageGroupTenant, tenantConf)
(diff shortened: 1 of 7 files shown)


cluster/cluster.go
```diff
@@ -289,6 +290,91 @@ func CreateDeploymentWithTenants(tenants []*StorageGroupTenant, sg *StorageGroup
...
+		}
+		ctx.Tenant = sgTenantResult.StorageGroupTenant.Tenant
+		// Get the credentials for a tenant
+		tenantConf, err := GetTenantConfig(tenant)
+		if err != nil {
+			ch <- err
+			return
+		}
+		// Unset Postgres notification before deleting tenant
+		err = unsetMinioConfigPostgresNotification(sgTenantResult.StorageGroupTenant, tenantConf)
```
dvaldivia, line 316: when you delete the storage, this should also get deleted right?

cluster/cluster.go
```diff
@@ -289,6 +290,91 @@ func CreateDeploymentWithTenants(tenants []*StorageGroupTenant, sg *StorageGroup
...
+		if err != nil {
+			log.Println(err)
+			ch <- err
+			return
+		}
+		// Wait after minios restart
+		ready := isMinioReadyRetry(ctx)
```
dvaldivia, line 323: I think this is an unnecessary delay

dvaldivia: If we don't unset the bucket notification configuration, we can just keep going with the deprovisioning

cesnietor:
Closing since it is outdated

PR #204, opened 2019-12-18
Global buckets

dvaldivia:
Enables global buckets by turning on logic on nginx if `SETUP_USE_GLOBAL_BUCKETS` is set to `true` during setup

### DIFF
--- a/cluster/buckets.go
+++ b/cluster/buckets.go
@@ -257,12 +257,6 @@ func streamBucketToTenantServices() chan *BucketToServiceResult {
 		}
 		defer rows.Close()
 
-		err = rows.Err()
-		if err != nil {
-			ch <- &BucketToServiceResult{Error: err}
-			return
-		}
-
 		for rows.Next() {
 			// Save the resulted query on the User struct
 			b2s := BucketToService{}
@@ -272,6 +266,12 @@ func streamBucketToTenantServices() chan *BucketToServiceResult {
 				return
 			}
 			ch <- &BucketToServiceResult{BucketToService: &b2s}
+		}
+
+		err = rows.Err()
+		if err != nil {
+			ch <- &BucketToServiceResult{Error: err}
+			return
 		}
 	}()
 	return ch
--- a/cluster/buckets.go
+++ b/cluster/buckets.go
@@ -257,6 +257,12 @@ func streamBucketToTenantServices() chan *BucketToServiceResult {
 		}
 		defer rows.Close()
 
+		err = rows.Err()
+		if err != nil {
+			ch <- &BucketToServiceResult{Error: err}
+			return
+		}
+
 		for rows.Next() {
 			// Save the resulted query on the User struct
 			b2s := BucketToService{}
@@ -305,7 +311,7 @@ func GetBucketUsageFromDB(ctx *Context, date time.Time) ([]*BucketMetric, error)
 					a.year,
 					a.month,
 					a.day,
-					greatest(0, (total_usage_average - previous_total_usage_average)) as daily_average_usage
+					greatest(0, (total_usage_average - previous_total_usage_average)) AS daily_average_usage
 				FROM(
 					SELECT 
 						a.year,
--- a/cluster/buckets.go
+++ b/cluster/buckets.go
@@ -17,6 +17,7 @@
 package cluster
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"log"
@@ -25,9 +26,10 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/v6"
+	"github.com/minio/minio-go/v6/pkg/policy"
 	"github.com/minio/minio/pkg/madmin"
+	uuid "github.com/satori/go.uuid"
 )
 
 type BucketAccess int32
@@ -37,6 +39,45 @@ const (
 	BucketPublic
 	BucketCustom
 )
+
+// MakeBucket will get the credentials for a given tenant and use the operator keys to create a bucket using minio-go
+// TODO: allow to spcify the user performing the action (like in the API/gRPC case)
+func MakeBucket(tenantShortname, bucketName string, accessType BucketAccess) error {
+	// validate bucket name
+	if bucketName != "" {
+		var re = regexp.MustCompile(`^[a-z0-9-]{3,}$`)
+		if !re.MatchString(bucketName) {
+			return errors.New("a valid bucket name is needed")
+		}
+	}
+
+	// Get tenant specific MinIO client
+	minioClient, err := newTenantMinioClient(nil, tenantShortname)
+	if err != nil {
+		return err
+	}
+
+	// make it so this timeouts after only 2 seconds
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+	// Create Bucket on tenant's MinIO
+	if err = minioClient.MakeBucketWithContext(timeoutCtx, bucketName, "us-east-1"); err != nil {
+		log.Println(err)
+		return tagErrorAsMinio(err)
+	}
+
+	if err = addMinioBucketNotification(minioClient, bucketName, "us-east-1"); err != nil {
+		log.Println(err)
+		return tagErrorAsMinio(err)
+	}
+
+	return SetBucketAccess(minioClient, bucketName, accessType)
+}
+
+type TenantBucketInfo struct {
+	Name   string
+	Access BucketAccess
+}
 
 func SetBucketAccess(minioClient *minio.Client, bucketName string, accessType BucketAccess) (err error) {
 	// Prepare policyJSON corresponding to the access type
@@ -67,41 +108,6 @@ func ChangeBucketAccess(tenantShortname, bucketName string, accessType BucketAcc
 	}
 
 	return SetBucketAccess(minioClient, bucketName, accessType)
-}
-
-// MakeBucket will get the credentials for a given tenant and use the operator keys to create a bucket using minio-go
-// TODO: allow to spcify the user performing the action (like in the API/gRPC case)
-func MakeBucket(tenantShortname, bucketName string, accessType BucketAccess) error {
-	// validate bucket name
-	if bucketName != "" {
-		var re = regexp.MustCompile(`^[a-z0-9-]{3,}$`)
-		if !re.MatchString(bucketName) {
-			return errors.New("a valid bucket name is needed")
-		}
-	}
-
-	// Get tenant specific MinIO client
-	minioClient, err := newTenantMinioClient(nil, tenantShortname)
-	if err != nil {
-		return err
-	}
-
-	// Create Bucket on tenant's MinIO
-	if err = minioClient.MakeBucket(bucketName, "us-east-1"); err != nil {
-		return err
-	}
-
-	if err = addMinioBucketNotification(minioClient, bucketName, "us-east-1"); err != nil {
-		log.Println(err)
-		return err
-	}
-
-	return SetBucketAccess(minioClient, bucketName, accessType)
-}
-
-type TenantBucketInfo struct {
-	Name   string
-	Access BucketAccess
 }
 
 // GetBucketAccess returns the access type for the given bucket name
@@ -143,10 +149,13 @@ func ListBuckets(tenantShortname string) ([]TenantBucketInfo, error) {
 		return []TenantBucketInfo{}, err
 	}
 
+	tCtx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+
 	var buckets []minio.BucketInfo
-	buckets, err = minioClient.ListBuckets()
+	buckets, err = minioClient.ListBucketsWithContext(tCtx)
 	if err != nil {
-		return []TenantBucketInfo{}, err
+		return []TenantBucketInfo{}, tagErrorAsMinio(err)
 	}
 
 	var (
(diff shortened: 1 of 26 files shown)


cluster/buckets.go
```diff
@@ -0,0 +1,266 @@
...
+		bucketAccess = BucketCustom
+	}
+
+	return bucketAccess, nil
+}
+
+// ListBuckets for the given tenant's short name
+func ListBuckets(tenantShortname string) ([]TenantBucketInfo, error) {
+	// Get tenant specific MinIO client
+	minioClient, err := newTenantMinioClient(nil, tenantShortname)
```
Alevsk: Should `newTenantMinioClient` be called using a context?

cluster/buckets.go
```diff
@@ -0,0 +1,266 @@
...
+		bucketInfos = append(bucketInfos, TenantBucketInfo{Name: bucket.Name, Access: accessType})
+	}
+
+	return bucketInfos, nil
+}
+
+// Deletes a bucket in the given tenant's MinIO
+func DeleteBucket(tenantShortname, bucket string) error {
+	// Get tenant specific MinIO client
+	minioClient, err := newTenantMinioClient(nil, tenantShortname)
```
Alevsk: Should newTenantMinioClient be called using a context?

cluster/buckets.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return err
+	}
+	// create the bucket registry
+	query :=
+		`INSERT INTO
+			buckets ("name","tenant_id")
+		VALUES
+			($1, $2)
+		ON CONFLICT DO NOTHING`
```
Alevsk, line 198: Do we want to avoid the bucket name `is unique` validation at the db level?

cluster/cluster.go
```diff
@@ -41,11 +41,10 @@ func getK8sConfig() *rest.Config {
...
-			BearerTokenFile: "some/file",
+			TLSClientConfig: rest.TLSClientConfig{Insecure: true},
+			APIPath:         "/",
+			BearerToken:     "eyJhbGciOiJSUzI1NiIsImtpZCI6InFETTJ6R21jMS1NRVpTOER0SnUwdVg1Q05XeDZLV2NKVTdMUnlsZWtUa28ifQ.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJkZWZhdWx0Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZWNyZXQubmFtZSI6ImRldi1zYS10b2tlbi14eGxuaiIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VydmljZS1hY2NvdW50Lm5hbWUiOiJkZXYtc2EiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC51aWQiOiJmZDVhMzRjNy0wZTkwLTQxNTctYmY0Zi02Yjg4MzIwYWIzMDgiLCJzdWIiOiJzeXN0ZW06c2VydmljZWFjY291bnQ6ZGVmYXVsdDpkZXYtc2EifQ.woZ6Bmkkw-BMV-_UX0Y-S_Lkb6H9zqKZX2aNhyy7valbYIZfIzrDqJYWV9q2SwCP20jBfdsDS40nDcMnHJPE5jZHkTajAV6eAnoq4EspRqORtLGFnVV-JR-okxtvhhQpsw5MdZacJk36ED6Hg8If5uTOF7VF5r70dP7WYBMFiZ3HSlJBnbu7QoTKFmbJ1MafsTQ2RBA37IJPkqi3OHvPadTux6UdMI8LlY7bLkZkaryYR36kwIzSqsYgsnefmm4eZkZzpCeyS9scm9lPjeyQTyCAhftlxfw8m_fsV0EDhmybZCjgJi4R49leJYkHdpnCSkubj87kJAbGMwvLhMhFFQ",
```
Alevsk, line 47: is this token be the same across all development environments? I mean the information inside the the jwt

```
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "default",
  "kubernetes.io/serviceaccount/secret.name": "dev-sa-token-xxlnj",
  "kubernetes.io/serviceaccount/service-account.name": "dev-sa",
  "kubernetes.io/serviceaccount/service-account.uid": "fd5a34c7-0e90-4157-bf4f-6b88320ab308",
  "sub": "system:serviceaccount:default:dev-sa"
}
```

cluster/nginx.go
```diff
@@ -228,15 +229,59 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
...
+			ch <- err
+			return
+		}
+		if globalBuckets.ValBool() {
+			nginxConfiguration = getGlobalBucketNamespaceConfiguration()
+		} else {
+			nginxConfiguration = getLocalBucketNamespaceConfiguration(ctx)
+		}
+
+		fmt.Println(nginxConfiguration)
```
Alevsk: Please change to `log.` if this is for debugging purposes

cluster/nginx.go
```diff
@@ -254,7 +299,7 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
 					location / {
 						 proxy_set_header Upgrade $http_upgrade;
 						 proxy_set_header Connection "upgrade";
-						 client_max_body_size 50M;
+						 client_max_body_size 0;
```
Alevsk, line 300: 0 means no limit in the body size?  if yes I think we should declare a maximum limit, even if this is big , to avoid potential DOS

cluster/nginx.go
```diff
@@ -294,31 +337,154 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
...
+
+			log_format  main  '$http_host - $remote_addr - $remote_user [$time_local] "$request" '
+																		'$status $body_bytes_sent "$http_referer" '
+																		'"$http_user_agent" "$http_x_forwarded_for"';
+				server {
+					access_log /var/log/nginx/access.log main;
+					location / {
+						 proxy_set_header Upgrade $http_upgrade;
+						 proxy_set_header Connection "upgrade";
+						 client_max_body_size 0;
```
Alevsk, line 426: 0 means no limit in the body size? if yes I think we should declare a maximum limit, even if this is big , to avoid potential DOS

cluster/nginx.go
```diff
@@ -294,31 +337,154 @@ func UpdateNginxConfiguration(ctx *Context) chan error {
...
+		if bucketRouteResult.Error != nil {
+			log.Println(bucketRouteResult.Error)
+			continue
 		}
+		bucketRoute := bucketRouteResult.BucketToService
+		serverBlock := fmt.Sprintf(`
+				server {
+					server_name %s.%s;
+					ignore_invalid_headers off;
+					client_max_body_size 0;
```
Alevsk, line 470: 0 means no limit in the body size? if yes I think we should declare a maximum limit, even if this is big , to avoid potential DOS

cluster/tenants.go
```diff
@@ -823,3 +677,76 @@ func StopTenantServers(sgt *StorageGroupTenantResult) error {
...
+	}
+	if globalBuckets.ValBool() {
+		// The instance the MinIO instance identifies as
+		tenantConfig["MINIO_PUBLIC_IPS"] = sgTenant.ServiceName
+		// Domain under all MinIO instances check for
+		tenantConfig["MINIO_DOMAIN"] = fmt.Sprintf("domain.m3,%s.s3", sgTenant.Tenant.ShortName)
+		tenantConfig["MINIO_ETCD_ENDPOINTS"] = "http://m3-etcd-cluster-client:2379"
+		tenantConfig["MINIO_ETCD_PATH_PREFIX"] = fmt.Sprintf("%s/", sgTenant.Tenant.ShortName)
+	}
+	// Build the confirm map
```
Alevsk: typo in comment

k8s/deployments/m3-deployment.yaml.example
```diff
@@ -77,7 +78,11 @@ spec:
             - name: MAIL_PASSWORD
               value: "<DEV_EMAIL_PASSWORD>"
             - name: APP_URL
-              value: localhost:3000
+              value: <APP_ADDRESS>
+            - name: S3_DOMAIN
+              value: <YOUR_DOMAIN>
+            - name: SETUP_USE_GLOBAL_BUCKETS
+              value: "true"
```
Alevsk, line 85: Do we want this configuration to be enabled by default always?

cluster/nginx.go
harshavardhana: That would be 5TiB single stream.

cluster/nginx.go
harshavardhana: Btw we don't need to add connection upgrade here that seems wrong @dvaldivia 

```
server {
 listen 80;
 server_name example.com;

 # To allow special characters in headers
 ignore_invalid_headers off;
 # Allow any size file to be uploaded.
 # Set to a value such as 1000m; to restrict file size to a specific value
 client_max_body_size 0;
 # To disable buffering
 proxy_buffering off;

 location / {
   proxy_set_header X-Real-IP $remote_addr;
   proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
   proxy_set_header X-Forwarded-Proto $scheme;
   proxy_set_header Host $http_host;

   proxy_connect_timeout  300;
   # Default is HTTP/1, keepalive is only enabled in HTTP/1.1
   proxy_http_version 1.1;
   proxy_set_header Connection "";

   proxy_pass http://localhost:9000; # If you are using docker-compose this would be the hostname i.e. minio
   # Health Check endpoint might go here. See https://www.nginx.com/resources/wiki/modules/healthcheck/
   # /minio/health/live;
 }
}
```

cluster/buckets.go
```diff
@@ -0,0 +1,266 @@
...
+		for rows.Next() {
+			// Save the resulted query on the User struct
+			b2s := BucketToService{}
+			err = rows.Scan(&b2s.Bucket, &b2s.Service, &b2s.ServicePort)
+			if err != nil {
+				ch <- &BucketToServiceResult{Error: err}
+				return
+			}
+			ch <- &BucketToServiceResult{BucketToService: &b2s}
+		}
```
cesnietor, line 269: please handle rows errors.
```
err = rows.Close()
	if err != nil {
		...
	}
```

cluster/service-accounts.go
```diff
@@ -654,3 +654,62 @@ func RemoveMinioServiceAccount(ctx *Context, serviceAccount *ServiceAccount) err
...
+				ak2s := AccessKeyToTenantShortName{
+					TenantShortName: tenantRes.Tenant.ShortName,
+				}
+				err = rows.Scan(&ak2s.AccessKey)
+				if err != nil {
+					ch <- &AccessKeyToTenantShortNameResult{Error: err}
+					return
+				}
+				ch <- &AccessKeyToTenantShortNameResult{AccessKeyToTenantShortName: &ak2s}
+			}
```
cesnietor, line 710: please handle rows errors.
```
err = rows.Close()
	if err != nil {
		...
	}
```

cluster/storage-cluster.go
```diff
@@ -520,3 +520,47 @@ func GetTenantStorageGroupByShortName(ctx *Context, tenantShortName string) chan
...
+			tenant := Tenant{}
+			tRes := TenantServiceResult{}
+			err = rows.Scan(&tenant.ID, &tenant.Name, &tenant.ShortName, &tRes.Service, &tRes.Port)
+			if err != nil {
+				ch <- TenantServiceResult{Error: err}
+				return
+			}
+			tRes.Tenant = &tenant
+			ch <- tRes
+		}
```
cesnietor, line 562: please handle rows errors.
```
err = rows.Close()
	if err != nil {
		...
	}
```

cluster/buckets.go
dvaldivia: No need, the internal `GetTenantStorageGroupByShortName` will work with or without a context

cluster/buckets.go
dvaldivia: yes, since we only get notified of creation

cluster/cluster.go
dvaldivia: it's a development token, it's actually ignored by the `client-go`

cluster/nginx.go
dvaldivia: forgot to remove this line

cluster/nginx.go
dvaldivia: but now we are passing files, we could set it to MinIO max which is 5TB

cluster/tenants.go
dvaldivia: nice catfch

k8s/deployments/m3-deployment.yaml.example
dvaldivia: for now, yes

cluster/nginx.go
dvaldivia: @harshavardhana this is for the portal, not for S3

cesnietor (changes requested):

cluster/buckets.go
dvaldivia: they are being returned for the caller to manage the error, since this is a streaming function it's up to the caller to decide what to do

cluster/storage-cluster.go
dvaldivia: they are being returned for the caller to manage the error, since this is a streaming function it's up to the caller to decide what to do

cluster/service-accounts.go
dvaldivia: they are being returned for the caller to manage the error, since this is a streaming function it's up to the caller to decide what to do

cluster/buckets.go
cesnietor: http://go-database-sql.org/retrieving.html 
We can do it similar to how it is here 
typo: It should be `rows.Err()` instead of rows.Close()
```
err = rows.Err()
if err != nil {
	...
}
```
At the end because scan could probably not trigger ab error.

cesnietor (changes requested):

cluster/buckets.go
dvaldivia: no need, internal function will access the DB without contextx

cesnietor (approved):

PR #209, opened 2019-12-19
Calculate Total month usage with the max usage during that month

cesnietor:
Also added Cost multiplier for tenant to calculate the cost by doing `total_usage * cost_multiplier`

### DIFF
--- a/cluster/buckets.go
+++ b/cluster/buckets.go
@@ -305,8 +305,7 @@ type BucketMetric struct {
 
 // GetTotalMonthBucketUsageFromDB get max total bucket usage of the month
 func GetTotalMonthBucketUsageFromDB(ctx *Context, date time.Time) (monthUsage uint64, err error) {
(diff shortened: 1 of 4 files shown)


cluster/buckets.go
```diff
@@ -303,8 +303,69 @@ type BucketMetric struct {
...
+		return 0, err
+	}
+	return cost, nil
+}
+
+// GetTotalMonthBucketUsageFromDB get max total bucket usage of the month
+func GetTotalMonthBucketUsageFromDB(ctx *Context, date time.Time) (monthUsage uint64, err error) {
+	// Select query doing MAX total_usage grouping by year and month
+	query := `SELECT 
+					MAX(s.total_usage) as total_monthly_usage
```
dvaldivia, line 334: Isn't it average?

dvaldivia (approved):

PR #217, opened 2019-12-20
Nginx Hot Reload

dvaldivia:
Makes some changes to Nginx so that whenever there's config maps updates, it downlaods the configuration and reloads the nginx without the need to restart the pod


harshavardhana:
Is this still a WIP ?

harshavardhana (approved):

PR #218, opened 2019-12-21
Wait for m3 tenant add, allow for cancellation

harshavardhana:
Adding tenant time is unpredictable in most
environments, allow for tenant delete during
cancellation in the middle handles any partial
creates.


Alevsk (approved):

harshavardhana:
With preprovisioning of MinIO this is not needed.

PR #224, opened 2019-12-22
Task Scheduler

dvaldivia:
Task Scheduler is designed to schedule jobs on the kubernetes cluster using jobs. This is intended for long running jobs such as provisioning a new tenant.

A new deployment `m3-scheduler` is added to the example `m3-deployment` so update your local file accordingly, this is because of the new configMap for environment variables `m3-env` which is shared among `m3`, `m3-scheduler` and all the scheduled jobs that run inside `m3`

So this PR can be reviewed while I finish the pre-provisioning of tenants I only added 2 tasks that can be scheduled, `fail` and `empty`, one fails and one suceeds properly.

### DIFF
--- a/cluster/scheduler.go
+++ b/cluster/scheduler.go
@@ -1,0 +1,253 @@
+// This file is part of MinIO Kubernetes Cloud
+// Copyright (c) 2019 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


cluster/scheduler.go
```diff
@@ -0,0 +1,253 @@
...
+func markTask(task *Task, newStatus TaskStatus) error {
+	// if we are marking the task as scheduled, set the schedule time field
+	extraField := ""
+	if newStatus == ScheduledTaskStatus {
+		extraField = ", scheduled_time = now()"
+	}
+	// build the query
+	query := fmt.Sprintf(
+		`UPDATE tasks 
+					SET status = $1 %s
```
Alevsk, line 114: vulnerable but not exploitable sqli

Alevsk (approved):

PR #225, opened 2019-12-22
Pre-Provision Tenants

dvaldivia:
Big change.

This introduces tenant pre-provisioning, which allocated a tenant with a randomized `short name` and gets created via `Task Scheduler` so it sits waiting for a `tenant add` action, to grab it, when it becomes taken, a new tenant is pre-provisioned, up to 16 per storage group.

Adding multiple storage groups will guarantee multiple tenants can be added quickly.

The old "short name"  identifier for a tenant was moved to the `domain` field

### DIFF
--- a/cluster/storage-cluster.go
+++ b/cluster/storage-cluster.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 
 	uuid "github.com/satori/go.uuid"
 )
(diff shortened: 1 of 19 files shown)


cluster/storage-cluster.go
```diff
@@ -571,3 +612,54 @@ func streamTenantService(maxChanSize int) chan TenantServiceResult {
...
+	total, err := totalNumberOfTenantInStorageGroup(&sg.ID)
+	if err != nil {
+		return err
+	}
+	if total >= maxNumberOfTenantsPerSg {
+		return errors.New("Max number of tenants on that storage group reached")
+	}
+
+	// pre-provision the first tenant of this storage group
+	provTenantName := strings.ToLower(RandomCharString(16))
```
Alevsk, line 627: Seems like now the tenant shortname is a random identifier, should we consider to migrate this field to uuid?

cluster/storage-cluster.go
```diff
@@ -571,3 +612,54 @@ func streamTenantService(maxChanSize int) chan TenantServiceResult {
...
+	// check if tenant name is available
+	for {
```
Alevsk, line 629: If we use uuid for shortname field then this for will be no necessary

cluster/tenants.go
```diff
@@ -35,6 +36,7 @@ type Tenant struct {
 	ID        uuid.UUID
 	Name      string
 	ShortName string
+	Domain    string
```
Alevsk, line 39: We have ID and ShortName bot being random generated strings, seems like domain is doing the same behavior that shortname used to do (tenant identifier defined by the user), why do we need the domain field now?

Alevsk (approved):

cluster/tenants.go
harshavardhana: Probably because of DNS requirements, since they are decoupled it is perhaps possible to use uuid.

cluster/storage-cluster.go
dvaldivia: uuid might not be url compatible, and could easy max out the max length of kubernetes names for things, so I think it's healthy to have a random character string that we control the size of

harshavardhana: Perhaps uuid is better

cluster/storage-cluster.go
dvaldivia: technically we don't need this check, but I decided to put it anyway :P

cluster/tenants.go
dvaldivia: ID is an identifier, short name will be renamed later, since it's meant to be an identifiable name we can use inside kubernetes that doesn't max out the name length constraints on k8s, domain is what the tenant used to elect before as short name, but since now we use that to track all the internals of a tenant, domain was added to deal with that

PR #228, opened 2019-12-23
Bug Adding First Admin

dvaldivia:


Alevsk (approved):

cluster/setup.go
```diff
@@ -514,10 +514,16 @@ func AddM3Admin(name, email string) error {
...
+		if err = apptCtx.Rollback(); err != nil {
+			log.Println(err)
+		}
 		log.Println("Error adding user:", err.Error())
 		return err
 	}
-	apptCtx.Commit()
+	// if no error, commit
+	if err = apptCtx.Commit(); err != nil {
+		log.Println(err)
```
harshavardhana, line 525: @dvaldivia don't you wish to return error here? instead of returning `nil` ?

PR #234, opened 2019-12-23
Start using env package to avoid repeated code

harshavardhana:


cluster/config.go
```diff
@@ -32,25 +33,17 @@ type hostConfigV9 struct {
...
+	return env.Get(s3Domain, "s3.localhost")
 }
 
 func getM3ContainerImage() string {
-	concreteM3Image := "minio/m3:dev"
-	if os.Getenv(m3Image) != "" {
-		concreteM3Image = os.Getenv(m3Image)
-	}
-	return concreteM3Image
+	return env.Get(m3Image, "minio/m3:edge")
```
cesnietor, line 40: I think this should be :dev cause all the yamls refer to that one.

harshavardhana: we don't have that container so its a bug to have :dev @cesnietor

cesnietor (approved):

dvaldivia (approved):

PR #243, opened 2019-12-28
Move all DB queries to Transactions withing Context

dvaldivia:
**Big change, need a lot of eyes**

With the current approach to managing database connections we are opening a db connection per tenant, this means that at any given point if we have `n` tenants we have `1+n` connections on any instance of m3 (m3, scheduler or a task) this quickly overruns the max number of connections supported by a regular postgres (100 on a 4GB ram instance)

To fix this we are moving to a connection model, similar to Django, where all the operations always happen within  a transaction, and the transaction is started at the begining of the request, since we already have the `cluster.Context` and we already keep transactions there, I just moved all non-transacitonal queries to use ctx transactions.

More over, the transactions are started when the context is started.

`db.GetInstance().Db` is deprecated and should be avoided

### DIFF
--- a/api/api-private-tenant-permissions.go
+++ b/api/api-private-tenant-permissions.go
@@ -54,7 +54,9 @@ func (ps *privateServer) TenantPermissionAdd(ctx context.Context, in *pb.TenantP
 	if err != nil {
 		return nil, status.New(codes.InvalidArgument, "Invalid tenant name").Err()
 	}
-	appCtx.Tenant = &tenant
+	if err := appCtx.SetTenant(&tenant); err != nil {
+		return nil, status.New(codes.Internal, "Internal error").Err()
+	}
 
 	if _, err := cluster.AddPermissionToDB(appCtx, in.Name, in.Description, effect, in.Resources, in.Actions); err != nil {
 		appCtx.Rollback()
@@ -79,7 +81,9 @@ func (ps *privateServer) TenantPermissionList(ctx context.Context, in *pb.Tenant
 		log.Println(err)
 		return nil, status.New(codes.InvalidArgument, "Invalid tenant name").Err()
 	}
-	appCtx.Tenant = &tenant
+	if err := appCtx.SetTenant(&tenant); err != nil {
+		return nil, status.New(codes.Internal, "Internal error").Err()
+	}
 	// perform actions
 	perms, err := cluster.ListPermissions(appCtx, in.Offset, in.Limit)
 	if err != nil {
@@ -133,7 +137,9 @@ func (ps *privateServer) TenantPermissionAssign(ctx context.Context, in *pb.Tena
 		log.Println(err)
 		return nil, status.New(codes.InvalidArgument, "Invalid tenant name").Err()
 	}
-	appCtx.Tenant = &tenant
+	if err := appCtx.SetTenant(&tenant); err != nil {
+		return nil, status.New(codes.Internal, "Internal error").Err()
+	}
 
 	// validate the permission
 	// get the permission to see if it's valid
--- a/api/api-private-tenant-permissions.go
+++ b/api/api-private-tenant-permissions.go
@@ -66,13 +66,13 @@ func (ps *privateServer) TenantPermissionList(ctx context.Context, in *pb.Tenant
 	appCtx, err := getContextIfValidTenant(ctx, in.Tenant)
 	if err != nil {
 		log.Println(err)
-		return nil, status.New(codes.Internal, "Internal error 1").Err()
+		return nil, status.New(codes.Internal, "Internal error").Err()
 	}
 	// perform actions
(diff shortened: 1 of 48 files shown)


cesnietor:
this happens right after running setup on kind cluster:
```
2019/12/30 01:58:24 pq: CREATE DATABASE cannot run inside a transaction block
2019/12/30 01:58:24 error connecting to database or reading migrations
2019/12/30 01:58:24 sql: Scan error on column index 0, name "current_schema": converting NULL to string is unsupported in line 0: SELECT CURRENT_SCHEMA()
2019/12/30 01:58:24 Adding the first admin
2019/12/30 01:58:26 Error adding user: pq: relation "admins" does not exist
2019/12/30 01:58:26 Could not store global bucket configuration. pq: current transaction is aborted, commands ignored until end of transaction block
2019/12/30 01:58:26 pq: current transaction is aborted, commands ignored until end of transaction block
2019/12/30 01:58:26 Starting m3 services...
2019/12/30 01:58:26 missing config
2019/12/30 01:58:26 pq: relation "configurations" does not exist
```

cesnietor:
when I try to login with two different users on the same tenant I get:
```
v2019/12/30 06:50:27 /m3.PublicAPI/Login
2019/12/30 06:50:27 pq: sorry, too many clients already
```

api/api-private-tenant-permissions.go
```diff
@@ -68,23 +63,16 @@ func (ps *privateServer) TenantPermissionAdd(ctx context.Context, in *pb.TenantP
 }
 
 func (ps *privateServer) TenantPermissionList(ctx context.Context, in *pb.TenantPermissionListRequest) (*pb.TenantPermissionListResponse, error) {
-	appCtx, err := cluster.NewEmptyContextWithGrpcContext(ctx)
+	appCtx, err := getContextIfValidTenant(ctx, in.Tenant)
 	if err != nil {
 		log.Println(err)
-		return nil, status.New(codes.Internal, "Internal error").Err()
+		return nil, status.New(codes.Internal, "Internal error 1").Err()
```
nitisht: This should be just "Internal Error" AFAICS

api/api-private-tenant-permissions.go
```diff
@@ -68,23 +63,16 @@ func (ps *privateServer) TenantPermissionAdd(ctx context.Context, in *pb.TenantP
...
-		log.Println(err)
-		return nil, status.New(codes.InvalidArgument, "Invalid tenant name").Err()
-	}
-	appCtx.Tenant = &tenant
 	// perform actions
 	perms, err := cluster.ListPermissions(appCtx, in.Offset, in.Limit)
 	if err != nil {
 		log.Println(err)
-		return nil, status.New(codes.Internal, "Internal error").Err()
+		return nil, status.New(codes.Internal, "Internal error 2").Err()
```
nitisht: This should be just "Internal Error" AFAICS

api/api-private-tenant.go
```diff
@@ -109,54 +109,50 @@ func (ps *privateServer) TenantAdd(in *pb.TenantAddRequest, stream pb.PrivateAPI
...
+		//	log.Println("Error getting tenants config", err)
+		//	return status.New(codes.Internal, "Error getting tenants config").Err()
+		//}
+
+		//// create minio postgres configuration for bucket notification
+		//err = cluster.SetMinioConfigPostgresNotification(sgt.StorageGroupTenant, tenantConf)
+		//if err != nil {
+		//	log.Println("Error setting tenant's minio postgres configuration", err)
+		//	return status.New(codes.Internal, "Error setting tenant's minio postgres configuration").Err()
+		//}
```
nitisht, line 138: can we remove the commented code

api/api-public-authentication-grpc.go
```diff
@@ -167,25 +182,27 @@ func (s *server) Login(ctx context.Context, in *pb.LoginRequest) (res *pb.LoginR
...
-		}
-		// if no error happened to this point commit transaction
-		err = appCtx.Commit()
-	}()
 	// Everything looks good, create session
-	session, err := cluster.CreateSession(appCtx, &user, &tenant)
+	session, err := cluster.CreateSession(appCtx, user, tenant)
+	//if err != nil {
+	//	return nil, status.New(codes.Internal, err.Error()).Err()
+	//}
```
nitisht, line 189: this can be removed as well

api/api-public-tenant-permissions.go
```diff
@@ -140,36 +135,28 @@ func (s *server) UpdatePermission(ctx context.Context, in *pb.UpdatePermissionRe
...
 	}
 
-	defer func() {
-		if err != nil {
-			appCtx.Rollback()
-			return
-		}
-	}()
-
 	// get permission
```
nitisht, line 138: No rollback now?

nitisht (changes requested):
few comments

PR #244, opened 2019-12-29
get buckets sizes from saved minio info during list buckets

cesnietor:
avoid calling minio twice during list buckets.


cluster/buckets.go
```diff
@@ -487,7 +515,7 @@ func GetDailyAvgBucketUsageFromDB(ctx *Context, date time.Time) ([]*BucketMetric
 // RecurrentTenantMetricsCalculation loop that calculates bucket usage metrics for all tenants and saves them on the db
 func RecurrentTenantMetricsCalculation() chan error {
 	// How often will this function run
-	ticker := time.NewTicker(6 * time.Hour)
```
harshavardhana: Calculating metrics for every 10seonds is a bad idea..

cesnietor: oh that was just for testing, bringing back to 6 hours

Alevsk (approved):

PR #248, opened 2019-12-30
Set MaxRetry to 1 to avoid retries in minio-go

harshavardhana:
- Use error wrapping whenever possible
- "DNS" style requests are not useful for
  admin client remove it.


cesnietor (approved):
LGTM tested creating tenant, buckets, and setting permissions.

PR #264, opened 2020-01-06
Add Bucket Error details

dvaldivia:


cluster/buckets.go
```diff
@@ -43,14 +43,16 @@ const (
...
 )
 
+var ErrInvalidBucketName = errors.New("invalid bucket name")
+
 // MakeBucket will get the credentials for a given tenant and use the operator keys to create a bucket using minio-go
 // TODO: allow to spcify the user performing the action (like in the API/gRPC case)
 func MakeBucket(ctx *Context, tenantShortname, bucketName string, accessType BucketAccess) error {
 	// validate bucket name
 	if bucketName != "" {
 		var re = regexp.MustCompile(`^[a-z0-9-]{3,}$`)
```
cesnietor: this need to change since this is not the proper rule for bucket names

dvaldivia: you mean the rule?

cesnietor: yes, https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

cesnietor (approved):

PR #265, opened 2020-01-07
vault_with_tls - enabling running vault with tls

Alevsk:


vault/Dockerfile
```diff
@@ -17,6 +17,7 @@ EXPOSE 8200
 
 COPY --from=0 /go/src/github.com/minio/m3-vault/m3-vault    .
 COPY --from=0 /go/src/github.com/minio/m3-vault/vault-config.json    .
+COPY --from=0 /go/src/github.com/minio/m3-vault/tls/    /tls/
```
dvaldivia, line 20: this one should be mounted via docker volumes directive

vault/tls/server.cert
```diff
@@ -0,0 +1,8 @@
+-----BEGIN CERTIFICATE-----
```
dvaldivia, line 1: is it possible to auto generate this per dev? this may expire at any given point

vault/cmd/m3-vault/main.go
```diff
@@ -181,17 +179,21 @@ func vaultInitAndUnseal() chan error {
...
-
-		//<-storeConfInK8s(rootToken)
-
 	}()
 	return doneCh
 }
 
 func isVaultReadyRetry(address string) (*vapi.Client, error) {
 	currentTries := 0
+	totalRetries, _ := strconv.Atoi(env.Get("TOTAL_INIT_RETRIES", "5"))
```
cesnietor, line 188: can you please add a case to handle error e.g.  `TOTAL_INIT_RETRIES -> "string"`
else the for will break badly if that happens, which could happen.

dvaldivia (approved):

PR #267, opened 2020-01-07
Vault without tls for local development

Alevsk:


harshavardhana:
@Alevsk do you want m3-vault to built always? on docker hub?

Alevsk:
> @Alevsk do you want m3-vault to built always? on docker hub?

m3-vault is intended to be used only for developers in their local k8s clusters, it would be nice to have always the latest m3-vault image on docker hub so we can pull it from there but if that add significant time to the build process we can skip it for now, currently in the `k8s/create-kind.sh` script we are building the image during the cluster setup process:

```
make --directory="../vault" k8sdev
...
...
kubectl apply -f deployments/m3-vault-deployment.yaml
```

cesnietor (approved):

PR #268, opened 2020-01-08
Moving KES configuration to env variables

Alevsk:


harshavardhana:
One suggestion why not follow a proper PR title, since we look at `git log` to be linear history it is always useful to keep meaningful git commit messages - there is a great example https://chris.beams.io/posts/git-commit/

Since we are now getting to a stage of stabilization perhaps we should make an effort to keep the PR titles readable.

Alevsk:
> One suggestion why not follow a proper PR title, since we look at `git log` to be linear history it is always useful to keep meaningful git commit messages - there is a great example https://chris.beams.io/posts/git-commit/
> 
> Since we are now getting to a stage of stabilization perhaps we should make an effort to keep the PR titles readable.

Thanks @harshavardhana I'll take a look

cluster/kes.go
```diff
@@ -162,17 +163,21 @@ path "kv/%s/*" {
 
 func getNewKesDeployment(deploymentName string, kesSecretsNames map[string]string) appsv1.Deployment {
 	kesReplicas := int32(1)
+	port, _ := strconv.Atoi(getKesRunningPort())
+	kesMTlsAuth := getkesMTlsAuth()
+	kesConfigPath := getKesConfigPath()
+	kesCommand := fmt.Sprintf("for i in {1..5}; do sleep 3; kes key create app-key -k && break || sleep 1; done & kes server --config=%s --mtls-auth=%s", kesConfigPath, kesMTlsAuth)
```
cesnietor, line 168: why do we loop 5 times  ?

Alevsk: We try 5 times to create the initial kes key, 5 its an arbitrary number, if we success the first time then we stop trying.

cesnietor (dismissed):

harshavardhana (changes requested):
you may have to `git commit --amend` locally with the title change `Moving KES configuration to env variables`

cluster/config.go
```diff
@@ -36,6 +36,18 @@ func getKesContainerImage() string {
 	return env.Get(kesImage, "minio/kes:latest")
 }
 
+func getKesRunningPort() string {
```
aead: Should we do input validation here?
Like checking that the port is a numder (uint16) and return an int instead of a string?

cluster/kes.go
```diff
@@ -162,17 +163,21 @@ path "kv/%s/*" {
 
 func getNewKesDeployment(deploymentName string, kesSecretsNames map[string]string) appsv1.Deployment {
 	kesReplicas := int32(1)
+	port, _ := strconv.Atoi(getKesRunningPort())
```
aead: Error handling?

cluster/kes.go
```diff
@@ -319,6 +324,7 @@ func createNewKesDeployment(clientset *kubernetes.Clientset, deploymentName stri
 
 func createNewKesService(clientset *kubernetes.Clientset, serviceName string) <-chan struct{} {
 	doneCh := make(chan struct{})
+	port, _ := strconv.Atoi(getKesRunningPort())
```
aead: Error handling?

cesnietor (changes requested):
can you add this commit in a separate pr since this has nothing to do with KES

Alevsk:
> can you add this commit in a separate pr since this has nothing to do with KES
> 
> I screwed sorry xd

done

dvaldivia (approved):

harshavardhana (approved):

cesnietor (approved):

PR #277, opened 2020-01-10
Deprovision of a tenant

cesnietor:
Removes all tenant's content but keeping the Services so that they can be reused for future tenants.

First it creates a task for being run inside a kubernetes job which will:

1. first move the mount folder to a provisional folder, then the provisional folder gets deleted and recreated. 
2. Once the folders are recreated the database schema gets deleted and recreated empty. 
3. Then we make the service available for other new tenants 
4. Restart MinIO servers so that they initialize in the new empty mount path.

Please take your time to test this.
To test it it is necessary to have a tenant already running then:
Disable Tenant
```
./m3 tenant disable <tenant_domain>
```
Delete Tenant
```
./m3 tenant delete <tenant_domain> --confirm
```

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -290,25 +290,14 @@ func CreateDeploymentWithTenants(tenants []*StorageGroupTenant, sg *StorageGroup
 }
 
 // DeprovisionTenantOnStorageGroup deletes the tenant from the storage group and deletes all tenant's data from disks
-func DeprovisionTenantOnStorageGroup(ctx *Context, tenant *Tenant, sg *StorageGroup) chan error {
+func DeprovisionTenantOnStorageGroup(ctx *Context, sgt *StorageGroupTenantResult) chan error {
 	ch := make(chan error)
 	go func() {
 		defer close(ch)
-		if tenant == nil || sg == nil {
-			ch <- errors.New("nil Tenant or StorageGroup passed")
-			return
-		}
-
-		sgTenantResult := <-GetTenantStorageGroupByShortName(ctx, tenant.ShortName)
-		if sgTenantResult.Error != nil {
-			ch <- sgTenantResult.Error
-			return
-		}
-
 		// start the jobs that create the tenant folder on each disk on each node of the storage group
 		var jobChs []chan error
 		// get a list of nodes on the cluster
-		nodes, err := GetNodesForStorageGroup(ctx, &sg.ID)
+		nodes, err := GetNodesForStorageGroup(ctx, &sgt.StorageGroupTenant.StorageGroup.ID)
 		if err != nil {
 			ch <- err
 			return
@@ -317,8 +306,9 @@ func DeprovisionTenantOnStorageGroup(ctx *Context, tenant *Tenant, sg *StorageGr
 			ch <- errors.New("Nodes not found to deprovision the tenant")
 			return
 		}
+
 		for _, sgNode := range nodes {
-			jobCh := DeleteTenantFolderInDisk(tenant, sg, sgNode)
+			jobCh := RecreateTenantFolderInDisk(sgt.StorageGroupTenant.Tenant, sgt.StorageGroupTenant.StorageGroup, sgNode)
 			jobChs = append(jobChs, jobCh)
 		}
 		// wait for all the jobs to complete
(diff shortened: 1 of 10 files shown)


cluster/cluster.go
```diff
@@ -482,7 +451,7 @@ func DeleteTenantFolderInDisk(tenant *Tenant, sg *StorageGroup, sgNode *StorageG
 
 			newFolderForDeletion := fmt.Sprintf("%s/%s-to-delete-%s", vol.MountPath, tenant.ShortName, randSringForDeletion)
 			// move current tenant path for one to be deleted and delete if afterwards
-			commands = append(commands, fmt.Sprintf(`mv -v %s/%s %s && rm -rv %s`, vol.MountPath, tenant.ShortName, newFolderForDeletion, newFolderForDeletion))
+			commands = append(commands, fmt.Sprintf(`mv -v %s/%s %s && rm -rv %s && mkdir -p %s/%s`, vol.MountPath, tenant.ShortName, newFolderForDeletion, newFolderForDeletion, vol.MountPath, tenant.ShortName))
```
harshavardhana: don't use `rm -rv` it may be too many files it's okay to just do `rm -rf`

cluster/cluster.go
```diff
@@ -482,7 +451,7 @@ func DeleteTenantFolderInDisk(tenant *Tenant, sg *StorageGroup, sgNode *StorageG
 
 			newFolderForDeletion := fmt.Sprintf("%s/%s-to-delete-%s", vol.MountPath, tenant.ShortName, randSringForDeletion)
```
harshavardhana: it is perhaps better to use a much more unique name so that avoids `mv` conflict perhaps stick a uuid into this directory.

cluster/cluster.go
harshavardhana: Also we don't need `mv -v` perhaps `mv -f` is better

cluster/minio.go
```diff
@@ -124,6 +124,19 @@ func stopMinioTenantServers(sgt *StorageGroupTenant, tenantConf *TenantConfigura
 	return nil
 }
 
+func restartMinioTenantServers(sgt *StorageGroupTenant, tenantConf *TenantConfiguration) error {
+	adminClient, pErr := NewAdminClient(sgt.HTTPAddress(false), tenantConf.AccessKey, tenantConf.SecretKey)
+	if pErr != nil {
+		return pErr.Cause
+	}
+	// Restart minios after setting configuration
```
harshavardhana: comment should be just `restart` there is no config set.

harshavardhana (changes requested):

cluster/cluster.go
Alevsk: Even if these values were auto generated (tenant shortname, etc) we need to validate all the parameters that goes into exec to minimize the possibility of command injection vulnerabilities via a previous sql injection, we can use regex for that https://github.com/minio/m3/commit/43b37eaba1784a8d1514dfc1959f9b53ea6bf704#diff-535fa68c8799a7d7e71f34e59b8c34faR59, please validate that provided vol.MountPath, tenant.ShortName, newFolderForDeletion, newFolderForDeletion, vol.MountPath, tenant.ShortName consist only of `[a-z]` characters

Alevsk (changes requested):

cluster/cluster.go
harshavardhana: Or avoid all of this altogether and use helpers like renameAll and removeAll written in cmd/os-reliable.go in `minio/minio` project @cesnietor

cesnietor: done

cluster/cluster.go
cesnietor: done

cluster/cluster.go
cesnietor: done

cluster/minio.go
cesnietor: done

cluster/cluster.go
cesnietor: since this is just a job that is running a simple cli command on an Ubuntu Image. I think we can keep it like that, else it would need to have a whole go project just removing a folder, I added a validation of the tenant shortname to prevent it from being more than a valid name.  let me know your thoughts.

Alevsk (approved):

harshavardhana (approved):

PR #285, opened 2020-01-15
Prevent Duplicate Permissions

dvaldivia:


cluster/permissions.go
```diff
@@ -902,3 +908,60 @@ func MapPermissionsToIDs(ctx *Context, permissions []string) (map[string]*uuid.U
...
+		}
+		if len(parts) > 1 {
+			r.Pattern = parts[1]
+		} else {
+			r.Pattern = "*"
+		}
+		completeResources = append(completeResources, r.String())
+	}
+
+	query := `SELECT p.id
```
harshavardhana, line 1008: Just to detect if the permissions are unique why do you need to call the database @dvaldivia ? or am I missing something here?

harshavardhana (changes requested):

cluster/permissions.go
dvaldivia: We keep track of the permissions on the DB so as we are adding a new one, we validate via query.

cesnietor: I think the issue is when assigning permissions with the same rules to  a Service Account, not the creation of a permission itself. I'm gonna do some tests with this pr.

cesnietor: yes, I think this doesn't fix the real issue which is there is the same effect on a certain bucket in two different permissions.
Screen Shot 2020-01-15 at 5 42 04 PM
 and gives :
 ```
MinIO: `policy.MarshalJSON`, duplicate actions [s3:*], resources [arn:aws:s3:::bucket1/*] found in statements { Allow [s3:*] [arn:aws:s3:::bucket1/*] []}, { Allow [s3:*] [arn:aws:s3:::bucket1/*] []}
```

this should be fixed during ServiceAccount creation and just filtering if the statements are the same

**Review by cesnietor — CHANGES_REQUESTED** [review]
when editing it still allows duplicates

File: cluster/permissions.go (hunk 1 above)
Comment (dvaldivia) [reply cluster/permissions.go:1008]: fixed

File: cluster/permissions.go [hunk 2]
```diff
@@ -389,19 +395,19 @@ func buildPermissionsForRows(ctx *Context, rows *sql.Rows) ([]*Permission, error
...
-	err = getResourcesForPermissions(ctx, permissionsHash)
+	err = getResourcesForPermissionsInTx(ctx, permissionsHash, inTx)
 	if err != nil {
 		return nil, err
 	}
 	return permissions, nil
 }
 
-func getActionsForPermissions(ctx *Context, permsMap map[uuid.UUID]*Permission) error {
+func getActionsForPermissionsInTx(ctx *Context, permsMap map[uuid.UUID]*Permission, inTx bool) error {
```
cesnietor: what is the main reason behind `inTx` ?  When should we use it? is it really needed?

dvaldivia: Some times we need to do queries in a transation and sometimes outside of it

cesnietor (approved):

dvaldivia:
@harshavardhana do you need any more changes to this PR?

harshavardhana (approved):

PR #288, opened 2020-01-16
avoid permissions and service accounts with the same name on creation

cesnietor:
since there might be permissions already on deployments I'm not adding a unique index on the db as a migration since if there are already existing duplicates the migration would fail.


dvaldivia:
We don't have a production environment yet, so we should introduce the DB constraint

dvaldivia (changes requested):
Can you add a unique index?

cesnietor:
after discussion, adding db constraint.

dvaldivia (approved):

PR #294, opened 2020-01-20
Adding support for custom ca certificates

Alevsk:
If required, we can load custom ca certificates into the folder with the same name,
certificates will be installed when building the mkube image.
This solves the problem of using a kms with a certificate issued by a custom ca.
Note that operator still need to create a configmap in k8s based on the ca certificate and
define the KMS_CA_CERT_CONFIG_MAP and KMS_CA_CERT_FILE_NAME variables so kes
can use the custom ca certificate


ca-certificates/example-ca-cert.crt
```diff
@@ -0,0 +1,30 @@
+-----BEGIN CERTIFICATE-----
```
Alevsk, line 1: This is an example ca certificate generated with https://github.com/FiloSottile/mkcert, adding just for git to preserve the `ca-certificates` folder, its just a public key, no harm or anything in having it in the repository

Alevsk:
As @aead I'll check if vault sdk supports the definition of a `custom ca certificate`, if that's possible we still need to mount the k8s volume but at least we don't have to modify the `Dockerfile`

dvaldivia (approved):

dvaldivia:
Please algo merge this on `orchestrator`

PR #295, opened 2020-01-20
Global Buckets: Fix delete bucket

dvaldivia:


cesnietor (changes requested):
When making a new bucket or deleting:
The serviceName that is being fed is an api not a valid service name so it throws that error of :
`error processing event sql: no rows in result set`
This is what it is throwing with a debug log of `serviceName`
```
2020/01/20 22:26:16 /m3.PublicAPI/MakeBucket - e6ae2b58-517d-47a1-9e6c-31f633e395cf
2020/01/20 22:26:16 serviceName: 10.106.189.5
2020/01/20 22:26:16 error processing event sql: no rows in result set
2020/01/20 22:26:16 serviceName: zbfv2wduo1iecyfr-sg-1
```

cesnietor (dismissed):
tested works fine

cesnietor (approved):

PR #298, opened 2020-01-20
Only delete a bucket if no permissions are using it

cesnietor:
Since we use postgres db to update permissions I leverage that by filtering permissions where a bucket is being used before deleting a bucket.

### DIFF
--- a/cluster/buckets.go
+++ b/cluster/buckets.go
@@ -184,8 +184,8 @@ func DeleteBucket(ctx *Context, bucket string) error {
 		return err
 	}
 	// Verify if bucket is being used within a permission.
-	// 	If bucket is being used, we don't allow the deletion.
-	//  The permission needs to be updated first.
+	//	If bucket is being used, we don't allow the deletion.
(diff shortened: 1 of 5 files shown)


cluster/buckets.go
```diff
@@ -175,17 +176,54 @@ func ListBuckets(tenantShortname string) ([]TenantBucketInfo, error) {
...
 
+// bucketInPermission returns wether a bucket is being used within a permission or not
+func bucketInPermission(ctx *Context, bucketName string) (bool, error) {
+	query := `
+		SELECT 
+			COUNT(*)
+		FROM 
+			permissions_resources
+		WHERE 
+		    bucket_name = $1`
```
dvaldivia: you could speed up this by introducing `LIMIT 1`

dvaldivia: and adding an index to the `bucket_name` field

cesnietor: makes sense. Done

dvaldivia (approved):

PR #303, opened 2020-01-25
Support custom Etcd repository

dvaldivia:


cluster/config.go
```diff
@@ -141,9 +141,29 @@ func getCliCommand() string {
...
+func getEtcdImage() string {
+	return env.Get(etcdImage, "quay.io/coreos/etcd:3.4.0")
+}
+
+func getEtcdImageRepository() string {
+	parts := strings.Split(getEtcdImage(), ":")
+	if len(parts) > 0 {
+		return parts[0]
+	}
+	return ""
```
Alevsk, line 160: should we return here `quay.io/coreos/etcd` instead of an empty string?

cluster/etcd.go
```diff
@@ -260,8 +260,9 @@ func getEtcdCRDDeployment(clusterName string) *unstructured.Unstructured {
 				"name": clusterName,
 			},
 			"spec": map[string]interface{}{
-				"size":    3,
-				"version": "3.4.0",
+				"size":       3,
+				"version":    getEtcdImageTag(),
+				"repository": getEtcdImageRepository(),
```
Alevsk, line 265: there's an scenario here in which we end with
```
version: 'latest'
repository: ' '
```
Do we support that?

cesnietor (approved):

PR #306, opened 2020-01-29
Improve error reported by set-password

dvaldivia:


cmd/m3/set-password.go
```diff
@@ -101,7 +105,19 @@ func setPassword(ctx *cli.Context) error {
 	})
 
 	if err != nil {
-		log.Println(err)
+		if e, ok := status.FromError(err); ok {
+			switch e.Code() {
+			case codes.PermissionDenied:
```
cesnietor, line 110: can you please add some comments for the permissionDenied? since this is only for the caller also aborted

cmd/m3/set-password.go
```diff
@@ -101,7 +105,19 @@ func setPassword(ctx *cli.Context) error {
...
+				log.Println(e.Message())
+			case codes.Internal:
+				log.Println("The server found an internal error")
+			case codes.Aborted:
+				log.Println("the connection Aborted the call")
+			default:
+				log.Printf("(%s) %s\n", e.Code(), e.Message())
```
cesnietor, line 117:
```suggestion
				log.Printf("(%d) %s\n", e.Code(), e.Message())
```
code comes as integer else it will print it like `%!s(int=<value>)`

cesnietor (changes requested):

cesnietor (approved):

PR #308, opened 2020-01-29
Version Public API

dvaldivia:
Adds an API to check the m3 version at `/api/v1/version` this returns whatever was set at compilation time.

For docker, this is either the current tag being built or the current git tag


Alevsk (dismissed):

version/version.go
```diff
@@ -0,0 +1,21 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cesnietor, line 15:
```suggestion
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

```

cesnietor (changes requested):

Alevsk (approved):

cesnietor (approved):

PR #311, opened 2020-01-31
Take Postgres out of M3 Setup

dvaldivia:


cesnietor (changes requested):
can you please update the README for development to mention what is needed now for having a dev environment up and running?

cesnietor (changes requested):
error on m3 scheduler 
```
2020/01/31 19:33:30 problem checking on the setup of m3
2020/01/31 19:33:30 m3 is not setup yet, will sleep and then crash peacefully
2020/01/31 19:34:00 Closing all connections after a panic
```

dvaldivia:
@cesnietor it's expected that scheduler will crash at least once when deploying if the tasks table is not ready

cesnietor (approved):

PR #334, opened 2020-02-07
Honoring token expiration time if idp is enabled

Alevsk:


cluster/admin-sessions.go
```diff
@@ -36,19 +36,30 @@ type AdminSession struct {
 	WhoAmI           string
 }
 
-func CreateAdminSession(ctx *Context, adminID *uuid.UUID) (*AdminSession, error) {
+func CreateAdminSession(ctx *Context, adminID *uuid.UUID, idpSession bool, expirationDate time.Time) (*AdminSession, error) {
```
cesnietor, line 39: Can this be probably two separate functions for each behavior?
Since if idpSession is false it won't consider the mandatory input parameter expirationDate.

dvaldivia (approved):

PR #339, opened 2020-02-12
Storing correct session information in db

Alevsk:
Mkube is generating the right jwt with the right expiration time but
was storing wrong information in the admin_sessions table


dvaldivia (approved):

cesnietor:
so does that mean that the postgres db may have different times than mkube?

dvaldivia:
yes, they are now controlled by the application, this is to mirror idp expirations

PR #341, opened 2020-02-12
Split portal and object storage nginx

dvaldivia:
This takes out portal proxy routing from the `nginx-resolver` so now two different services need to be exposed to access the portal and the object storage service:

`nginx-resolver` will take all the object storage traffic meant for MinIO
`portal-proxy` will take all the requests relevant to the portal and it's APIs


cesnietor (changes requested):
Please update the Readme with those changes.

dvaldivia:
@cesnietor done

cesnietor (approved):

PR #345, opened 2020-02-20
add creation date on listBuckets public api

cesnietor:


cluster/buckets.go
```diff
@@ -38,6 +38,14 @@ import (
 
 type BucketAccess int32
 
+type BucketInfo struct {
+	// Name of the bucket.
+	Name   string
+	Access BucketAccess
+	// Date and time when the bucket was created.
```
harshavardhana, line 45: Does this struct not need Size?

harshavardhana (changes requested):

cluster/buckets.go
cesnietor: Size is considered on the public api struct response and is built on `api-public-bucket-grpc.go` `ListBuckets` since `ListBucketsWithContext` call of minio-go also doesn't include the size. This struct is similar on how minio has it on `object-api-datatypes.go`.

dvaldivia (dismissed):

Alevsk (approved):

harshavardhana (approved):

PR #346, opened 2020-02-25
getAppURL() was returning the wrong value

Alevsk:
`getAppURL()` now returns by default `http://localhost/` instead of
`http://$S3_DOMAIN/`, users can specify APP_URL with `https` and it will
be honored

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -133,8 +133,7 @@ func getMaxNumberOfTenantsPerSg() int {
 
 // AppURL returns the main application url
 func getAppURL() string {
-	appDomain := "localhost"
-	return env.Get("APP_URL", fmt.Sprintf("http://%s", appDomain))
+	return env.Get("APP_URL", "http://localhost")
 }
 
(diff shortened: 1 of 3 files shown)


cluster/config.go
```diff
@@ -133,7 +133,7 @@ func getMaxNumberOfTenantsPerSg() int {
 
 // AppURL returns the main application url
 func getAppURL() string {
-	appDomain := getS3Domain()
+	appDomain := "localhost"
 	return env.Get("APP_URL", fmt.Sprintf("http://%s", appDomain))
```
cesnietor:
```suggestion
	return env.Get("APP_URL", "http://localhost")
```
just a suggestion to avoid having a variable that behaves as a constant

cesnietor:
```suggestion
	return env.Get("APP_URL", "http://localhost")
```
just a suggestion to avoid having a variable that behaves as a constant

cesnietor (approved):

PR #348, opened 2020-02-25
Portal service refactor

Alevsk:


k8s/create-kind.sh
```diff
@@ -91,9 +95,12 @@ kubectl apply -f deployments/postgres-dev.yaml
 kubectl apply -f deployments/etcd-dev.yaml
 # install prometheus
 kubectl apply -f deployments/prometheus-dev.yaml
-kubectl apply -f deployments/m3-portal-frontend-deployment.yaml
-kubectl apply -f deployments/m3-vault-deployment.yaml
-kubectl apply -f deployments/portal-proxy-deployment.yaml
+# install rest of the services
+kubectl apply -f deployments/m3-vault.yaml
```
dvaldivia, line 99: let's comment out all vault instalation

portal-ui/public/index.html
```diff
@@ -1,47 +1 @@
-<!DOCTYPE html>
```
dvaldivia, line 1: Why not keep the formatted index.html? seems easier to manage and develop

portal-ui/public/precache-manifest.d1c8f4292065174e51a3d57ce7578ff7.js
```diff
@@ -0,0 +1,30 @@
+self.__precacheManifest = (self.__precacheManifest || []).concat([
```
dvaldivia, line 1: I think you checked in the compiled UI. Let's not do that, as the UI is built inside the container

dvaldivia (changes requested):

PR #358, opened 2020-03-12
MCS: Add List Users

dvaldivia:
Also fixes the copyright on all files


mcs/restapi/users.go
```diff
@@ -0,0 +1,104 @@
...
+
+// Interface implementation
+//
+// Define the structure of a minIO Client and define the functions that are actually used
+// from minIO api.
+type adminClient struct {
+	client *madmin.AdminClient
+}
+
+// implements minio.ListBucketsWithContext(ctx)
```
cesnietor: wrong comment

mcs/restapi/users.go
```diff
@@ -0,0 +1,104 @@
...
+			Policy:    user.PolicyName,
+			MemberOf:  user.MemberOf,
+		}
+		users = append(users, bucketElem)
+	}
+
+	return users, nil
+}
+
+// getListBucketsResponse performs listBuckets() and serializes it to the handler's output
```
cesnietor: wrong comment

mcs/restapi/admin_handlers_test.go
```diff
@@ -0,0 +1,89 @@
...
+	userMap, err := listUsers(adminClient)
+	if err != nil {
+		t.Errorf("Failed on %s:, error occurred: %s", function, err.Error())
+	}
+	// verify length of users is correct
+	assert.Equal(len(mockUserMap), len(userMap), fmt.Sprintf("Failed on %s: length of user's lists is not the same", function))
+
+	for _, b := range userMap {
+		assert.Contains(mockUserMap, b.AccessKey)
+		assert.Equal(string(mockUserMap[b.AccessKey].Status), b.Status)
```
cesnietor, line 78: could you please also add the validation of `b.PolicyName` and probably changing the `status` of one to `disabled` to be different and ensure that both are tested against each other. And also if possible validate `MemberOf` probably using `
`assert. ElementsMatch()`

Thanks for adding tests 👍

cesnietor (changes requested):

mcs/restapi/admin_handlers_test.go
dvaldivia: done, we don't have calls to update states on users yet, as soon as those are added I'll be putting a test for that

cesnietor (approved):
LGTM

PR #362, opened 2020-03-18
MCS add admin api for list, add and remove groups

cesnietor:
Also some files were renamed and reorganized.
Includes tests.


mcs/restapi/admin_groups.go
```diff
@@ -0,0 +1,168 @@
+// This file is part of MinIO Kubernetes Cloud
+// Copyright (c) 2019 MinIO, Inc.
```
dvaldivia: The year should be 2020

dvaldivia (changes requested):

Alevsk (dismissed):

mcs/restapi/admin_groups.go
cesnietor: done

dvaldivia (approved):

PR #363, opened 2020-03-19
MCS add updateGroup and getGroupInfo admin api

cesnietor:
update allows to update the members by either adding members or deleting members with a simple PUT operation.


mcs/restapi/admin_groups.go
```diff
@@ -166,3 +217,114 @@ func getRemoveGroupResponse(params admin_api.RemoveGroupParams) error {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func setGroupStatus(client MinioAdmin, group, status string) error {
+	var setStatus madmin.GroupStatus
+	switch status {
+	case "enabled":
```
dvaldivia, line 262: this is a good opportunity for a constant or an enum

dvaldivia (approved):

mcs/restapi/admin_groups.go
cesnietor: yes I thought about that mhmh but not sure if restrict it since minio might change the status

cesnietor: will do if necessary in future

PR #365, opened 2020-03-25
Mcs admin config api list and get config

cesnietor:
List configurations will be done with : `/api/v1/configs`:
Response:
```
{
    "configurations": [
        {
            "description": "label the location of the server",
            "key": "region"
        },
        {
            "description": "define object level redundancy",
            "key": "storage_class"
        },
        {
            "description": "add caching storage tier",
            "key": "cache"
        },
        {
            "description": "enable server side compression of objects",
            "key": "compression"
        },
        {
            "description": "federate multiple clusters for IAM and Bucket DNS",
            "key": "etcd"
        },
        {
            "description": "enable OpenID SSO support",
            "key": "identity_openid"
        },
        {
            "description": "enable LDAP SSO support",
            "key": "identity_ldap"
        },
        {
            "description": "enable external OPA for policy enforcement",
            "key": "policy_opa"
        },
        {
            "description": "enable external HashiCorp Vault key management service",
            "key": "kms_vault"
        },
        {
            "description": "enable external MinIO key encryption service",
            "key": "kms_kes"
        },
        {
            "description": "send server logs to webhook endpoints",
            "key": "logger_webhook"
        },
        {
            "description": "send audit logs to webhook endpoints",
            "key": "audit_webhook"
        },
        {
            "description": "publish bucket notifications to webhook endpoints",
            "key": "notify_webhook"
        },
        {
            "description": "publish bucket notifications to AMQP endpoints",
            "key": "notify_amqp"
        },
        {
            "description": "publish bucket notifications to Kafka endpoints",
            "key": "notify_kafka"
        },
        {
            "description": "publish bucket notifications to MQTT endpoints",
            "key": "notify_mqtt"
        },
        {
            "description": "publish bucket notifications to NATS endpoints",
            "key": "notify_nats"
        },
        {
            "description": "publish bucket notifications to NSQ endpoints",
            "key": "notify_nsq"
        },
        {
            "description": "publish bucket notifications to MySQL databases",
            "key": "notify_mysql"
        },
        {
            "description": "publish bucket notifications to Postgres databases",
            "key": "notify_postgres"
        },
        {
            "description": "publish bucket notifications to Elasticsearch endpoints",
            "key": "notify_elasticsearch"
        },
        {
            "description": "publish bucket notifications to Redis datastores",
            "key": "notify_redis"
        }
    ]
}
```
Will be different per Configuration.
For get specific config : `/api/v1/configs/{name}`:
Response:
```
{
    "keyvalue": [
        {
            "key": "enable",
            "value": "off"
        },
        {
            "key": "format",
            "value": "namespace"
        },
        {
            "key": "connection_string"
        },
        {
            "key": "table"
        },
        {
            "key": "host"
        },
        {
            "key": "port"
        },
        {
            "key": "username"
        },
        {
            "key": "password"
        },
        {
            "key": "database"
        },
        {
            "key": "queue_dir"
        },
        {
            "key": "queue_limit",
            "value": "0"
        }
    ],
    "name": "notify_postgres"
}
```


dvaldivia:
@cesnietor is it possible to return the type for the expected values?

cesnietor:
> @cesnietor is it possible to return the type for the expected values?

good question, let me check cause the call that we do only return the key and value in strings. But let me see if we can get that somewhere cause it is just on documentation right now: https://docs.min.io/docs/minio-bucket-notification-guide.html

dvaldivia (approved):

mcs/restapi/admin_config.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			return admin_api.NewConfigInfoDefault(500).WithPayload(&models.Error{Code: 500, Message: swag.String(err.Error())})
+		}
+		return admin_api.NewConfigInfoOK().WithPayload(config)
+	})
+
+}
+
+// listConfig gets all configurations' names and their descriptions
+func listConfig(client MinioAdmin) (madmin.Help, error) {
```
Alevsk, line 52: it would be better here to return `*madmin.Help` instead of `madmin.Help`, that way you can return `nil` instead of an empty `madmin.Help{}` struct, you can do that since on line `72` if this function returns an error you are returning `nil` for `*models.ListConfigResponse` anyway

mcs/restapi/admin_config.go
```diff
@@ -0,0 +1,133 @@
...
+		configDescs = append(configDescs, desc)
+	}
+	listGroupsResponse := &models.ListConfigResponse{
+		Configurations: configDescs,
+	}
+	return listGroupsResponse, nil
+}
+
+// getConfig gets the key values for a defined configuration
+func getConfig(client MinioAdmin, name string) (madmin.Target, error) {
```
Alevsk, line 92: Similar to previous comment, return type should be `*madmin.Target` instead of `madmin.Target` to avoid the `madmin.Target{}` struct

mcs/restapi/admin_config.go
```diff
@@ -0,0 +1,133 @@
...
+	config, err := client.getConfigKV(name)
+	if err != nil {
+		return madmin.Target{}, err
+	}
+	// config comes as an array []madmin.Target
+	if len(config) > 0 {
+		// return Key Values, first element contains info
+		return config[0], nil
+	}
+	return madmin.Target{}, nil
```
Alevsk, line 102: if you change `madmin.Target` to `*madmin.Target` this should be  &madmin.Target{}

mcs/restapi/admin_config.go
```diff
@@ -0,0 +1,133 @@
...
+		log.Println("error listing configurations:", err)
+		return nil, err
+	}
+
+	// serialize output
+	var confkv []*models.ConfigurationKV
+	confName := params.Name
+	for _, kv := range configTarget.KVS {
+		confkv = append(confkv, &models.ConfigurationKV{Key: kv.Key, Value: kv.Value})
+	}
```
Alevsk, line 127: I think this should be inside the `getConfig` function since `getConfigResponse` should be only responsible to populate the values of `&models.Configuration` and not compute them

mcs/restapi/admin_groups.go
```diff
@@ -72,17 +72,12 @@ func registerGroupsHandlers(api *operations.McsAPI) {
 }
 
 // listGroups calls MinIO server to list all groups names present on the server.
-func listGroups(client MinioAdmin) (groups []string, err error) {
+func listGroups(client MinioAdmin) ([]string, error) {
```
Alevsk, line 75: same here, it can be possible to return `[]*string` that allow you to return `nil` instead of  an empty `[]string{}` when there was an error `err`

mcs/restapi/admin_groups.go
```diff
@@ -72,17 +72,12 @@ func registerGroupsHandlers(api *operations.McsAPI) {
...
 	groupList, err := client.listGroups()
 	if err != nil {
-		return groups, err
+		return []string{}, err
```
Alevsk, line 78: this should be `return nil, err`

PR #366, opened 2020-03-25
Endpoints for info and set Policy

Alevsk:


mcs/restapi/admin_policies.go
```diff
@@ -211,3 +220,73 @@ func getAddPolicyResponse(params *models.AddPolicyRequest) (*models.Policy, erro
...
+		log.Println("error getting  group info:", err)
+		return nil, err
+	}
+	return policy, nil
+}
+
+// setPolicy() calls MinIO server to assign policy to a group or user.
+func setPolicy(client MinioAdmin, name, entityName, entityType string) error {
+	isGroup := false
+	if entityType == "group" {
```
cesnietor, line 263: what are the valid entity types? we can use here similar to what we do on Bucket Access an Enum:
By doing that the server itself will enforce only the valid types when trying to send invalid ones.
```
bucketAccess:
    type: string
    enum:
      - PRIVATE
      - PUBLIC
      - CUSTOM
    default: PRIVATE
  bucket:
    type: object
    required:
      - name
    properties:
      name:
        type: string
        minLength: 3
      size:
        type: integer
        format: int64
      access:
        $ref: "#/definitions/bucketAccess"
      creation_date:
        type: string
```

mcs/restapi/admin_policies_test.go
```diff
@@ -0,0 +1,281 @@
...
+func (ac adminClientMock) removePolicy(name string) error {
+	return minioRemovePolicyMock(name)
+}
+
+// mock function of addPolicy()
+func (ac adminClientMock) addPolicy(name, policy string) error {
+	return minioAddPolicyMock(name, policy)
+}
+
+// mock function setGroupStatus()
```
cesnietor: wrong comments

mcs/restapi/admin_policies_test.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil
+	}
+	// Test-1 : setPolicy() set policy to user
+	function := "setPolicy()"
+	err := setPolicy(adminClient, policyName, entityName, entityObject)
+	if err != nil {
+		t.Errorf("Failed on %s:, error occurred: %s", function, err.Error())
+	}
+	// Test-2 : setPolicy() set policy to group
+	entityObject = "group"
```
cesnietor: if you change the definition of the entityObject then on testing we can use the` models.Group` for instance definition

mcs/restapi/admin_policies_test.go
```diff
@@ -0,0 +1,281 @@
...
+		assert.Equal("error", err.Error())
+	}
+	// Test-4 : setPolicy() set policy to group and get error
+	entityObject = "group"
+	minioSetPolicyMock = func(policyName, entityName string, isGroup bool) error {
+		return errors.New("error")
+	}
+	if err := setPolicy(adminClient, policyName, entityName, entityObject); assert.Error(err) {
+		assert.Equal("error", err.Error())
+	}
```
cesnietor, line 280: A test using invalid entityObject? what should happen there? should we return an error `invalid entity object`?

cesnietor (changes requested):

mcs/restapi/admin_policies_test.go
Alevsk: This will not work, since that validation happen outside of the `setPolicy` function

cesnietor (approved):

PR #392, opened 2020-05-19
Cluster APIs Create and List

dvaldivia:
Removed all the files related to the previous strategy around having mkube be a k8s controller.
Added swagger
Added Create Cluster API
Added List Clusters API


Alevsk (approved):

cmd/m3/server.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	api := operations.NewM3API(swaggerSpec)
+	server := restapi.NewServer(api)
+	defer server.Shutdown()
+
+	parser := flags.NewParser(server, flags.Default)
+	parser.ShortDescription = "MinIO Console Server"
```
Alevsk:
```suggestion
	parser.ShortDescription = "Mkube"
```

Alevsk (changes requested):

restapi/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+// The alphabet of random character string. Each character must be unique.
+//
+// The RandomCharString implementation requires that: 256 / len(letters) is a natural numbers.
+// For example: 256 / 64 = 4. However, 5 > 256/62 > 4 and therefore we must not use a alphabet
+// of 62 characters.
+// The reason is that if 256 / len(letters) is not a natural number then certain characters become
+// more likely then others.
+const letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ012345"
+
```
cesnietor, line 38: you can import this from mcs/pkg/auth/utils to avoid duplicated functionality

cmd/m3/server.go
```diff
@@ -0,0 +1,140 @@
...
+	restapi.Hostname = ctx.String("host")
+	restapi.Port = fmt.Sprintf("%v", ctx.Int("port"))
+
+	tlsCertificatePath := ctx.String("tls-certificate")
+	tlsCertificateKeyPath := ctx.String("tls-key")
+
+	if tlsCertificatePath != "" && tlsCertificateKeyPath != "" {
+		server.TLSCertificate = flags.Filename(tlsCertificatePath)
+		server.TLSCertificateKey = flags.Filename(tlsCertificateKeyPath)
+		// If TLS certificates are provided enforce the HTTPS schema, meaning mcs will redirect
```
Alevsk, line 123:
```suggestion
		// If TLS certificates are provided enforce the HTTPS schema, meaning mkube will redirect
```

restapi/clusters.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+		return admin_api.NewCreateClusterCreated()
+	})
+}
+
+func getClusterCreatedResponse(params admin_api.CreateClusterParams) error {
+
+	if params.Body.Name == nil || *params.Body.Name == "" {
+		return errors.New("Name cannot be empty")
+	}
```
cesnietor: this is not needed, since it is already validated before getting here. you already declared it as `required:true`

restapi/clusters.go
```diff
@@ -0,0 +1,259 @@
...
+
+	var re = regexp.MustCompile(`^[a-z0-9-]{3,63}$`)
+	if !re.MatchString(*params.Body.Name) {
+		return errors.New("Invalid name. Must not contain spaces or non alpha-numerical characters")
+	}
```
cesnietor: you can do this on the definition of the name on the yaml like e.g.:
```
ssn:
  type: string
  pattern: '^\d{3}-\d{2}-\d{4}$'
```

restapi/clusters.go
```diff
@@ -0,0 +1,259 @@
...
+		minImg, err := getLatestMinIOImage()
+		if err == ErrCantDetermineMinIOImage {
+			return errors.New("can't determine the MinIO Version. please specify one")
+		} else if err != nil {
+			return err
+		}
+		minioImage = *minImg
+	}
+
+	if params.Body.VolumeConfiguration == nil {
```
cesnietor: if this is required, set it on the yaml

restapi/clusters.go
```diff
@@ -0,0 +1,259 @@
...
+		return errors.New("Please specify a volume configuration")
+	}
+
+	if params.Body.VolumeConfiguration.Size == "" {
```
cesnietor: if this is required, set it on the yaml

cesnietor (commented):
Also we should consider to add some atomicity cause if it fails somewhere after creating  e.g. secrets, those will be there flying around.

cesnietor:
Also this pr seems to do more than just  the `create cluster api` you can probably add more description on the commit.

cmd/m3/server.go
dvaldivia: done

restapi/clusters.go
dvaldivia: it's not required, objects if not set on the payload are received as null, this parameter is optional.

restapi/clusters.go
dvaldivia: mm, I guess this os is ineed required :P

restapi/clusters.go
cesnietor: why are you returning an error if it is not received? that makes it required?

dvaldivia: hahaha you are correct, I'm removing all this checks and only keeping them for the optional parameters

restapi/clusters.go
dvaldivia: done

restapi/clusters.go
dvaldivia: done

restapi/utils.go
dvaldivia: let's see if there's more functions eventually that we need from mcs to justify the dependency

Alevsk (approved):

PR #397, opened 2020-05-23
Start Mirror for mkube

Alevsk:
This PR adds support for start the mirroring process between a source
and a target minio cluster via the `/mirror`
endpoint, this endpoint will used
`OperatorV1().MirrorInstances(currentNamespace).Create` using the user
provided parameters: source host, target host, source URL and target URL.

Test using minio play instance:

```
curl --location --request POST 'http://localhost:8787/api/v1/mirror' \
--header 'Content-Type: application/json' \
--header 'Content-Type: text/plain' \
--data-raw '{
    "host_source":
"https://Q3AM3UQ867SPQQA43P2F:zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG@play.min.io",
    "host_target":
"https://Q3AM3UQ867SPQQA43P2F:zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG@play.min.io",
    "source": "source/alevsk",
    "target": "target/alevsk2",
    "mirror_flags": ["--overwrite"]
}'
```
The endpoint will return something like:

```
{
    "mirror_id": "mc-mirror-xbjq2dz0nq"
}
```

`mirror_id` will be used in the future to track the progress of a mirror
job.


cluster/config.go
```diff
@@ -87,3 +98,88 @@ func getKmsCACertFileName() string {
...
+	matches := re.FindAllStringSubmatch(string(body), 1)
+	for i := range matches {
+		release := matches[i][1]
+		dockerImage := fmt.Sprintf("minio/mc:%s", release)
+		return &dockerImage, nil
+	}
+	return nil, errCantDetermineMCImage
+}
+
+var latestMCImage, errLatestMCImage = getLatestMCImage()
```
cesnietor, line 173: can't this be inside GetMCImage? else you are making it global

restapi/mirror.go
```diff
@@ -0,0 +1,128 @@
...
+	// if mc image version is not provided we obtain one from m3 configuration
+	if mcImage == "" {
+		mcImg, err := cluster.GetMCImage()
+		if err != nil {
+			return nil, err
+		}
+		mcImage = *mcImg
+	}
+
+	if hostSource == "" || hostTarget == "" || sourceURL == "" || targetURL == "" {
```
cesnietor, line 76: should we log something here? are those params required? probably we can add that restriction on the yaml, if already defined there as required then this code is probably not needed.

restapi/mirror.go
```diff
@@ -0,0 +1,128 @@
...
+	// mirrorInstance will contains all the necessary credentials
+	mirrorInstance := operator.MirrorInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: instanceName,
+		},
+		Spec: operator.MirrorInstanceSpec{
+			Image: mcImage,
+			Env: []corev1.EnvVar{
+				{
+					Name:  "MC_HOST_source",
```
cesnietor, line 89: should this be in capital letters?

restapi/mirror.go
```diff
@@ -0,0 +1,128 @@
...
+					Value: hostSource,
+				},
+				{
+					Name:  "MC_HOST_target",
```
cesnietor, line 93: should this be in capital letters?

cluster/config.go
Alevsk: This is intentional, `var latestMCImage, errLatestMCImage = getLatestMCImage()` will be run only once when mkube starts, `getLatestMCImage()` does an expensive http.GET operation to `https://dl.min.io/` if `getLatestMCImage()` its moved inside `GetMCImage()` then this request will trigger whenever we use that function

restapi/mirror.go
Alevsk: i defined this parameters as required in the yaml, ill check if there's a way to tell swagger this variables should not be empty strings otherwise we need to check manually none of these are empty variables.

cesnietor: oh got it

cluster/config.go
cesnietor: got it

restapi/mirror.go
Alevsk: good catch, i think it should be all capital letters but right now its supported like that for minio-operator https://github.com/minio/minio-operator/blob/master/examples/mirrorinstance.yaml#L23 i think ill open an issue there

restapi/mirror.go
Alevsk: same as above

dvaldivia (approved):

PR #400, opened 2020-06-03
Delete Tenant Handle

dvaldivia:


restapi/tenants.go
```diff
@@ -63,6 +63,30 @@ func registerTenantHandlers(api *operations.M3API) {
...
+}
+
+func getDeleteTenantResponse(params admin_api.DeleteTenantParams) error {
+	opClient, err := operatorClientset.NewForConfig(cluster.GetK8sConfig())
+	if err != nil {
+		return err
+	}
+	currentNamespace := cluster.GetNs()
+
+	err = opClient.OperatorV1().MinIOInstances(currentNamespace).Delete(context.Background(), params.Name, metav1.DeleteOptions{})
```
Alevsk: can you implement this function call inside the `restapi/operator_client.go` ? so we can use it in future testing.

Alevsk (changes requested):

cesnietor (changes requested):
please add tests

Alevsk (dismissed):

restapi/tenants.go
```diff
@@ -63,6 +63,38 @@ func registerTenantHandlers(api *operations.M3API) {
...
+		}
+		return admin_api.NewTenantInfoOK()
+
+	})
+}
+
+// getDeleteTenantResponse gets the output of deleting a minio instance
+func getDeleteTenantResponse(params admin_api.DeleteTenantParams) error {
+	opClientClientSet, err := operatorClientset.NewForConfig(cluster.GetK8sConfig())
+	if err != nil {
```
cesnietor, line 83: can you add logs for the errors here and on the call of `deleteTenantAction()`

cesnietor (changes requested):

cesnietor (approved):

restapi/tenants.go
dvaldivia: we only need to log once the same error, done

PR #409, opened 2020-06-11
Set memory size when requesting MinioInstance on tenant creation

cesnietor:
Memory size is now set to 16Gi by default or can be changed using `M3_TENANT_MEMORY_SIZE` env variable.
To test this you need to create a new tenant and when getting the minioinstances on your k8s env the description of the minioInstance should have the request like e.g.:
```
"resources": {
                    "requests": {
                        "memory": "16Gi",
                        "storage": "1Gi"
                    }
                },
```

### DIFF
--- a/restapi/config.go
+++ b/restapi/config.go
@@ -38,6 +38,10 @@ var TLSPort = "8443"
 // TLSRedirect m3 tls redirect rule
 var TLSRedirect = "off"
 
+// defaultTenantMemorySize default value used
+// when generating minioInstance request
+var defaultTenantMemorySize = "16Gi"
+
 // GetHostname gets m3 hostname set on env variable,
 // default one or defined on run command
 func GetHostname() string {
@@ -68,4 +72,10 @@ func GetSSLPort() int {
 		port = 9443
 	}
 	return port
+}
+
+// getTenantMemorySize Memory size value to be used when generating the
+// MinioInstance request
+func getTenantMemorySize() string {
+	return env.Get(M3TenantMemorySize, defaultTenantMemorySize)
 }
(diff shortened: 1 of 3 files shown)


restapi/config.go
```diff
@@ -38,6 +38,10 @@ var TLSPort = "8443"
 // TLSRedirect m3 tls redirect rule
 var TLSRedirect = "off"
 
+// defaultTenantMemorySize default value used
+// when generating minioInstance request
+var defaultTenantMemorySize = "4Gi"
```
harshavardhana: I think lets move this to 16Gi and also storage to 1Ti by default we should choose sane deployment defaults, no one will deploy us with 1Gi disk.

harshavardhana (changes requested):

restapi/config.go
cesnietor: I'll change the capacity to 16Gi then, Storage is an input from the user.

cesnietor: done @harshavardhana

harshavardhana (approved):

harshavardhana:
>                         "storage": "1Gi"

If they don't choose anything does it pick 1Gi? @cesnietor

cesnietor:
> > ```
> >                     "storage": "1Gi"
> > ```
> 
> If they don't choose anything does it pick 1Gi? @cesnietor

if not sent, it will fail on operator, @dvaldivia introduced that validation on this pr:
https://github.com/minio/minio-operator/commit/b7ca5dce0cbf1dda01b491e372b4edc0f913c373

bexsoft (approved):
LGTM

PR #413, opened 2020-06-20
Change tenants api to be namespace scope

cesnietor:
To list tenants of all namespaces and add a new tenant use GET and POST `/api/v1/tenants`.
To list tenants from a namespace use GET on `api/v1/namespaces/<ns-name>/tenants`.
To get/delete a tenant on a namespace use GET/DELETE on `api/v1/namespaces/<ns-name>/tenants/<tenant-name>`.


dvaldivia (approved):

bexsoft (approved):
LGTM & works fine

PR #415, opened 2020-06-30
GKE Integration Support

dvaldivia:
These are all the little tweaks we added for the demo last week


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM golang:1.14.0
```
harshavardhana, line 1:
```suggestion
FROM golang:1.14
```

harshavardhana (changes requested):

Alevsk (approved):

Dockerfile
dvaldivia: done

harshavardhana (approved):

